PR #18, opened 2026-03-15
feat(ffprobe): implement Probe() using go-astiav CGo bindings

### LINKED ISSUE #3: feat(ffprobe): ffprobe wrapper package
## Summary

Implement `pkg/ffprobe` — a thin, idiomatic Go wrapper around `go-astiav` that returns a structured `MediaInfo` type used by workflow steps to make transcoding decisions.

## Acceptance Criteria

- [x] **Given** a valid local media file path, **when** `ffprobe.Probe(ctx, path)` is called, **then** it returns a `MediaInfo` value with no error.
- [x] **Given** the returned `MediaInfo`, **when** inspecting it, **then** it exposes: container format, duration, overall bitrate, and a slice of streams each containing codec name, codec type (video/audio/subtitle), and for video streams the width, height, and frame rate.
- [x] **Given** a path to a file that does not exist, **when** `ffprobe.Probe(ctx, path)` is called, **then** it returns a non-nil error.
- [x] **Given** a path to a file that is not a media file, **when** `ffprobe.Probe(ctx, path)` is called, **then** it returns a non-nil error.
- [x] **Given** a cancelled context, **when** `ffprobe.Probe(ctx, path)` is called, **then** it returns promptly with a context error.

## Technical Context

- Depends on: #1
- Library: `github.com/asticode/go-astiav v0.40.0` (Cgo bindings for FFmpeg 8 `libavformat`/`libavcodec`)
- `MediaInfo` should be a package-owned struct (not a direct re-export of the library type) so callers are insulated from upstream API changes
- CGO must be enabled at build time; FFmpeg 8 shared libraries must be present at runtime (no external binary required)
- Context cancellation is implemented via `astiav.NewIOInterrupter()` — set on the format context before `OpenInput`; check the interrupted flag after errors to return the context error

## Non-Goals

- No writing or modifying media files
- No hardware detection (that belongs in `pkg/ffmpeg`)

solidDoWant:
Fixes #3

## Summary
- Implements `Probe(ctx context.Context, path string) (MediaInfo, error)` in `pkg/ffprobe`
- Uses `go-astiav` CGo bindings to open media containers and inspect streams
- Context cancellation wired through `astiav.NewIOInterrupter()`
- `MediaInfo` exposes: container format, duration, overall bitrate, and a `[]StreamInfo` slice
- `StreamInfo` exposes: codec name, codec type, and width/height/frame-rate for video streams

## Test plan
- [x] `TestProbe_ValidFile` — probes `testdata/video.mp4`, asserts no error and all MediaInfo fields populated
- [x] `TestProbe_NonExistentFile` — probes a non-existent path, asserts error returned
- [x] `TestProbe_NonMediaFile` — probes this test file (Go source), asserts error returned
- [x] `TestProbe_CancelledContext` — cancels context before call, asserts `context.Canceled` returned
- [x] `make test` passes (with `-race`)
- [x] `make lint` passes (0 issues)


pkg/ffprobe/ffprobe.go
```diff
@@ -1,2 +1,103 @@
...
+	Format   string
+	Duration time.Duration
+	BitRate  int64
+	Streams  []StreamInfo
+}
+
+// StreamInfo holds per-stream metadata.
+type StreamInfo struct {
+	CodecName string
+	CodecType string  // "video", "audio", "subtitle", etc.
```
solidDoWant: This should probably be an "enum" type with constant values for comparison, if possible

pkg/ffprobe/ffprobe.go
```diff
@@ -1,2 +1,103 @@
...
+	Width     int     // non-zero for video streams
+	Height    int     // non-zero for video streams
```
solidDoWant: Field needs to include a unit (e.g. Px/Pixels)

pkg/ffprobe/ffprobe.go
```diff
@@ -1,2 +1,103 @@
...
+	FrameRate float64 // non-zero for video streams
```
solidDoWant: Field needs to include a unit (e.g. FramesPerSecond/similar)

pkg/ffprobe/ffprobe.go
```diff
@@ -1,2 +1,103 @@
...
+}
+
+// Probe opens the media file at path, reads its container and stream metadata,
+// and returns a populated MediaInfo. It returns a non-nil error if the file
+// does not exist, is not a recognized media format, or if ctx is cancelled.
+func Probe(ctx context.Context, path string) (MediaInfo, error) {
```
solidDoWant: Maybe return a pointer instead of a concrete struct, so that errors can return a nil value. What do you think?

pkg/ffprobe/ffprobe.go
```diff
@@ -1,2 +1,103 @@
...
+	defer fc.Free()
+
+	// Set up IO interrupter so FFmpeg respects context cancellation.
+	interrupter := astiav.NewIOInterrupter()
+	defer interrupter.Free()
+	fc.SetIOInterrupter(interrupter)
+
+	// Watch for context cancellation in the background.
+	watchDone := make(chan struct{})
+	defer close(watchDone)
```
solidDoWant, line 66: should this be called from within the goroutine below instead of holding on to it until this outer function exits?

pkg/ffprobe/ffprobe.go
```diff
@@ -1,2 +1,103 @@
...
+		case <-ctx.Done():
+			interrupter.Interrupt()
+		case <-watchDone:
+		}
+	}()
+
+	// Bail early if context is already cancelled.
+	if err := ctx.Err(); err != nil {
+		return MediaInfo{}, err
+	}
```
solidDoWant: Why is this check needed and why here? This smells racey, but I might be wrong.

pkg/ffprobe/ffprobe.go
```diff
@@ -1,2 +1,103 @@
...
+
+	// Open input file.
+	if err := fc.OpenInput(path, nil, nil); err != nil {
+		if interrupter.Interrupted() {
```
solidDoWant, line 78: Does something need to be done to ensure that the interrupter is ALWAYS interrupted when the function exits? For example, in this case if `interrupter.Interrupted() == false`, then it will not be interrupted prior to freeing it. I don't know if ffmpeg requries this or not.

pkg/ffprobe/ffprobe.go
```diff
@@ -1,2 +1,103 @@
...
+	"time"
+
+	"github.com/asticode/go-astiav"
+)
+
+// MediaInfo holds the top-level metadata for a media container.
+type MediaInfo struct {
+	Format   string
+	Duration time.Duration
+	BitRate  int64
```
solidDoWant: Needs unit suffix (e.g. BitsPerSecond)

pkg/ffprobe/ffprobe.go
```diff
@@ -1,2 +1,103 @@
...
+		}
+		return MediaInfo{}, fmt.Errorf("ffprobe: finding stream info for %q: %w", path, err)
+	}
+
+	info := MediaInfo{
+		Duration: time.Duration(fc.Duration()) * time.Microsecond,
+		BitRate:  fc.BitRate(),
+	}
+
+	if f := fc.InputFormat(); f != nil {
```
solidDoWant: don't use single letter vars here or elsewhere except for indexes. Also don't use short (e.g. cp, si) variable names, use something more descriptive

pkg/ffprobe/ffprobe.go
```diff
@@ -1,2 +1,103 @@
...
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/asticode/go-astiav"
+)
+
+// MediaInfo holds the top-level metadata for a media container.
```
solidDoWant: Based on what libav returns, do you think any other fields would be useful here or in StreamInfo?

pkg/ffprobe/ffprobe_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/solidDoWant/media-processor/pkg/ffprobe"
+)
+
+// testdataPath returns the absolute path to a file under pkg/ffprobe/testdata/.
+func testdataPath(name string) string {
+	_, thisFile, _, _ := runtime.Caller(0)
+	return filepath.Join(filepath.Dir(thisFile), "testdata", name)
+}
```
solidDoWant: Is there a more idiomatic way of doing this? I don't really like the `runtime` call here.

pkg/ffprobe/ffprobe_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+func TestProbe_ValidFile(t *testing.T) {
+	info, err := ffprobe.Probe(context.Background(), testdataPath("video.mp4"))
```
solidDoWant: use the testing context. You've hit this issue a few times on a few issues, so you should probably add this to claude.md as well.

pkg/ffprobe/ffprobe_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/solidDoWant/media-processor/pkg/ffprobe"
+)
+
+// testdataPath returns the absolute path to a file under pkg/ffprobe/testdata/.
+func testdataPath(name string) string {
+	_, thisFile, _, _ := runtime.Caller(0)
+	return filepath.Join(filepath.Dir(thisFile), "testdata", name)
+}
+
+func TestProbe_ValidFile(t *testing.T) {
```
solidDoWant, line 15: This should assert the actual file values, not just non-empty/positive/etc. Query the file via ffprobe if needed to get the test data once before hard-coding it here.

pkg/ffprobe/ffprobe_test.go
```diff
@@ -0,0 +1,77 @@
...
+	// Must have at least one stream.
+	require.NotEmpty(t, info.Streams)
+
+	// Every stream must have a codec name and type.
+	for _, s := range info.Streams {
+		assert.NotEmpty(t, s.CodecName)
+		assert.NotEmpty(t, s.CodecType)
+	}
+
+	// Must have at least one video stream with width, height, and frame rate.
```
solidDoWant: After getting the actual file info, add tests for other streams as well.

pkg/ffprobe/testdata/video.mp4
solidDoWant, line 1: This file is the first few seconds of Big Buck Bunny. Be sure to include an require license notices in this directory, as well as a link to the source (not the upstream library, but the actual movie source).

pkg/ffprobe/ffprobe.go
solidDoWant: Added  as a named type with constants (, , , , , ) for comparison.

pkg/ffprobe/ffprobe.go
solidDoWant: Renamed to WidthPixels and HeightPixels.

pkg/ffprobe/ffprobe.go
solidDoWant: Renamed to FramesPerSecond.

pkg/ffprobe/ffprobe.go
solidDoWant: Changed to return *MediaInfo so errors return nil.

pkg/ffprobe/ffprobe.go
solidDoWant: After moving Free() to the goroutine, it is always called after the goroutine exits. FFmpeg does not require the callback to be in an interrupted state before freeing — av_free() just releases the allocation.

pkg/ffprobe/ffprobe.go
solidDoWant: Renamed to BitRateBitsPerSecond (on both MediaInfo and StreamInfo).

pkg/ffprobe/ffprobe_test.go
solidDoWant: Switched to a plain relative path (filepath.Join("testdata", "video.mp4")). Go test always runs with CWD set to the package directory, so runtime.Caller is unnecessary.

pkg/ffprobe/testdata/video.mp4
solidDoWant: Added testdata/README.md with CC BY 3.0 attribution, source URL, and copyright notice.

pkg/ffprobe/ffprobe.go
solidDoWant: Moved interrupter.Free() inside the goroutine — it is now called after the select exits (whether via ctx.Done() or watchDone closing). Removed defer interrupter.Free() from the outer function.

pkg/ffprobe/ffprobe.go
solidDoWant: Removed. The IOInterrupter already handles cancellation — when ctx is cancelled the goroutine calls Interrupt() before OpenInput/FindStreamInfo runs, so they return an interrupted error. The explicit ctx.Err() check was redundant and introduced a race (context could cancel in the window between the check and OpenInput).

pkg/ffprobe/ffprobe.go
solidDoWant: Renamed all single-letter and abbreviated variables: s -> stream, cp -> codecParams, si -> streamInfo, f -> inputFormat.

pkg/ffprobe/ffprobe.go
solidDoWant: Added Tags map[string]string to MediaInfo (container-level metadata from the format context dictionary) and to each stream. Also added per-stream BitRateBitsPerSecond, and for audio streams: AudioSampleRateHz and AudioChannelCount. These seemed like the most broadly useful additions from what libav exposes.

pkg/ffprobe/ffprobe_test.go
solidDoWant: Switched to t.Context() throughout. Also added a rule to CLAUDE.md: always use t.Context() instead of context.Background() in tests so the context is automatically cancelled when the test finishes.

pkg/ffprobe/ffprobe_test.go
solidDoWant: Replaced all non-empty/positive assertions with exact expected values obtained by running ffprobe on the file: format "mov,mp4,m4a,3gp,3g2,mj2", duration 5013333us, bitrate 607664, video stream h264 320x180 24fps 441324bps, audio stream aac 48kHz stereo 161052bps. Container tags (title, artist) also asserted.

pkg/ffprobe/ffprobe_test.go
solidDoWant: Added full assertions for both the audio stream (aac, 48kHz, 2-channel stereo, 161052 bps) and verified the video stream's zero audio fields. Both streams are now fully covered.

pkg/ffprobe/ffprobe.go
```diff
@@ -10,40 +10,58 @@ import (
...
+
+// MediaInfo holds top-level metadata for a media container.
 type MediaInfo struct {
-	Format   string
-	Duration time.Duration
-	BitRate  int64
-	Streams  []StreamInfo
+	Format               string
+	Duration             time.Duration
+	BitRateBitsPerSecond int64
```
solidDoWant: Change to just BitsPerSecond

pkg/ffprobe/ffprobe.go
```diff
@@ -10,40 +10,58 @@ import (
...
 // StreamInfo holds per-stream metadata.
 type StreamInfo struct {
-	CodecName string
-	CodecType string  // "video", "audio", "subtitle", etc.
-	Width     int     // non-zero for video streams
-	Height    int     // non-zero for video streams
-	FrameRate float64 // non-zero for video streams
+	CodecName            string
+	CodecType            CodecType
+	BitRateBitsPerSecond int64
```
solidDoWant: Change to just BitsPerSecond

pkg/ffprobe/ffprobe.go
```diff
@@ -52,52 +70,73 @@ func Probe(ctx context.Context, path string) (MediaInfo, error) {
...
+	if err := formatContext.FindStreamInfo(nil); err != nil {
 		if interrupter.Interrupted() {
-			return MediaInfo{}, ctx.Err()
+			return nil, ctx.Err()
 		}
-		return MediaInfo{}, fmt.Errorf("ffprobe: finding stream info for %q: %w", path, err)
+		return nil, fmt.Errorf("ffprobe: finding stream info for %q: %w", path, err)
 	}
 
 	info := MediaInfo{
```
solidDoWant: should this be `&MediaInfo` instead of just `MediaInfo`?

pkg/ffprobe/ffprobe.go
solidDoWant: Renamed to `BitsPerSecond`.

pkg/ffprobe/ffprobe.go
solidDoWant: Renamed to `BitsPerSecond`.

pkg/ffprobe/ffprobe.go
solidDoWant: Changed to `info := &MediaInfo{...}` so the pointer is allocated directly.

PR #20, opened 2026-03-18
feat(medialib): Radarr and Sonarr client helpers

### LINKED ISSUE #5: feat(medialib): Radarr and Sonarr client helpers
## Summary

Implement `pkg/medialib` — Go helpers wrapping `golift/starr` for the specific Radarr and Sonarr operations needed by the media processing workflows.

## Acceptance Criteria

- [ ] **Given** a valid Radarr base URL and API key, **when** `radarr.GetMovieByFilePath(ctx, path)` is called with a path that Radarr knows about, **then** it returns a `Movie` value containing at minimum: the Radarr movie ID, title, and year.
- [ ] **Given** `radarr.GetMovieByFilePath(ctx, path)` is called with a path Radarr does not know about, **then** it returns a sentinel `ErrNotFound` error.
- [ ] **Given** a valid Radarr movie ID, **when** `radarr.RefreshMovie(ctx, id)` is called, **then** it triggers a Radarr library rescan for that movie and returns nil on success.
- [ ] **Given** a valid Sonarr base URL and API key, **when** `sonarr.GetEpisodeByFilePath(ctx, path)` is called with a path that Sonarr knows about, **then** it returns an `Episode` value containing at minimum: the Sonarr episode ID, series title, season number, and episode number.
- [ ] **Given** `sonarr.GetEpisodeByFilePath(ctx, path)` is called with a path Sonarr does not know about, **then** it returns `ErrNotFound`.
- [ ] **Given** a valid Sonarr series ID, **when** `sonarr.RefreshSeries(ctx, id)` is called, **then** it triggers a Sonarr library rescan for that series and returns nil on success.
- [ ] **Given** an unreachable base URL, **when** any client function is called, **then** it returns a non-nil error within the context deadline.

## Technical Context

- Depends on: #1
- Library: `golift.io/starr`, `golift.io/starr/radarr`, `golift.io/starr/sonarr`
- Radarr config sourced from env vars: `RADARR_URL`, `RADARR_API_KEY`
- Sonarr config sourced from env vars: `SONARR_URL`, `SONARR_API_KEY`
- `GetMovieByFilePath` / `GetEpisodeByFilePath` will need to list all records and match by file path, as the *Arr APIs do not offer a direct path-lookup endpoint
- Expose `ErrNotFound` as a package-level sentinel so callers can distinguish "not in library" from other errors

## Non-Goals

- No adding movies/episodes to Radarr/Sonarr
- No managing quality profiles or download clients

solidDoWant:
## Summary
- Add `pkg/medialib/radarr` and `pkg/medialib/sonarr` packages wrapping `golift.io/starr`
- Define `MovieLibrary` and `EpisodeLibrary` interfaces in `pkg/medialib` for future backend swappability
- Optional `LocalPathPrefix`/`RemotePathPrefix` config per client to handle differing mount points between the worker and *Arr services (populated by the `cmd/` layer from env vars — library does not read env vars directly)

## Acceptance criteria
- [x] `radarr.GetMovieByFilePath` returns `Movie` for known path
- [x] `radarr.GetMovieByFilePath` returns `ErrNotFound` for unknown path
- [x] `radarr.RefreshMovie` returns nil on success
- [x] `sonarr.GetEpisodeByFilePath` returns `Episode` for known path
- [x] `sonarr.GetEpisodeByFilePath` returns `ErrNotFound` for unknown path
- [x] `sonarr.RefreshSeries` returns nil on success
- [x] Unreachable URL returns non-nil error within context deadline

## Test plan
- All tests use `httptest.NewServer` stubs — no real Radarr/Sonarr required
- `make test` and `make lint` pass cleanly

Fixes #5

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -76,6 +76,7 @@ require (
 	golang.org/x/sync v0.20.0 // indirect
 	golang.org/x/sys v0.42.0 // indirect
 	golang.org/x/text v0.35.0 // indirect
+	golift.io/starr v1.3.1 // indirect
 	google.golang.org/genproto/googleapis/api v0.0.0-20260209200024-4cfbd4190f57 // indirect
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20260226221140-a57be14db171 // indirect
 	google.golang.org/grpc v1.79.2 // indirect
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/asticode/go-astiav v0.40.0
 	github.com/hatchet-dev/hatchet v0.79.43
 	github.com/stretchr/testify v1.11.1
+	golift.io/starr v1.3.1
 	gopkg.in/yaml.v3 v3.0.1
 )
 
@@ -76,7 +77,6 @@ require (
 	golang.org/x/sync v0.20.0 // indirect
 	golang.org/x/sys v0.42.0 // indirect
 	golang.org/x/text v0.35.0 // indirect
-	golift.io/starr v1.3.1 // indirect
 	google.golang.org/genproto/googleapis/api v0.0.0-20260209200024-4cfbd4190f57 // indirect
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20260226221140-a57be14db171 // indirect
 	google.golang.org/grpc v1.79.2 // indirect
--- a/pkg/medialib/medialib.go
+++ b/pkg/medialib/medialib.go
@@ -1,2 +1,37 @@
-// Package medialib provides higher-level abstractions over ffmpeg and ffprobe.
+// Package medialib provides higher-level abstractions for media library services.
 package medialib
+
+import (
+	"context"
+	"errors"
+)
+
+// ErrNotFound is returned when a media item is not found in the library.
+var ErrNotFound = errors.New("not found in library")
+
+// Movie represents a movie entry in a movie library service.
+type Movie struct {
+	ID    int64
+	Title string
+	Year  int
+}
+
+// Episode represents an episode entry in a TV library service.
+type Episode struct {
+	ID            int64
+	SeriesTitle   string
+	SeasonNumber  int
+	EpisodeNumber int
+}
+
+// MovieLibrary provides operations for movie media items.
+type MovieLibrary interface {
+	GetMovieByFilePath(ctx context.Context, path string) (Movie, error)
+	RefreshMovie(ctx context.Context, id int64) error
+}
+
+// EpisodeLibrary provides operations for episode media items.
+type EpisodeLibrary interface {
+	GetEpisodeByFilePath(ctx context.Context, path string) (Episode, error)
+	RefreshSeries(ctx context.Context, seriesID int64) error
+}
--- a/pkg/medialib/medialib.go
+++ b/pkg/medialib/medialib.go
@@ -11,6 +11,7 @@ var ErrNotFound = errors.New("not found in library")
 
 // Movie represents a movie entry in a movie library service.
 type Movie struct {
+	// ID is the internal database ID assigned by the backing movie library service.
 	ID    int64
 	Title string
 	Year  int
@@ -18,6 +19,7 @@ type Movie struct {
 
 // Episode represents an episode entry in a TV library service.
 type Episode struct {
+	// ID is the internal database ID assigned by the backing TV library service.
 	ID            int64
 	SeriesTitle   string
 	SeasonNumber  int
--- a/pkg/medialib/medialib.go
+++ b/pkg/medialib/medialib.go
@@ -20,7 +20,10 @@ type Movie struct {
 // Episode represents an episode entry in a TV library service.
 type Episode struct {
 	// ID is the internal database ID assigned by the backing TV library service.
-	ID            int64
+	ID int64
+	// SeriesID is the internal database ID of the series this episode belongs to.
+	// Use this with EpisodeLibrary.RefreshSeries to trigger a library rescan.
+	SeriesID      int64
 	SeriesTitle   string
 	SeasonNumber  int
 	EpisodeNumber int
--- a/pkg/medialib/radarr/client.go
+++ b/pkg/medialib/radarr/client.go
@@ -1,0 +1,81 @@
+// Package radarr provides a Radarr client implementing medialib.MovieLibrary.
+package radarr
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"golift.io/starr"
+	radarrlib "golift.io/starr/radarr"
+
+	"github.com/solidDoWant/media-processor/pkg/medialib"
+)
+
+// Compile-time assertion that *Client implements medialib.MovieLibrary.
+var _ medialib.MovieLibrary = (*Client)(nil)
+
+// Config holds the configuration for a Radarr client.
+type Config struct {
+	URL    string
+	APIKey string
+	// LocalPathPrefix and RemotePathPrefix enable optional path translation.
+	// When set, paths starting with LocalPathPrefix are rewritten to use
+	// RemotePathPrefix before being matched against Radarr's stored paths.
+	// This handles cases where the worker and Radarr use different mount points
+	// for the same storage.
+	LocalPathPrefix  string
+	RemotePathPrefix string
+}
+
+// Client is a Radarr client implementing medialib.MovieLibrary.
+type Client struct {
+	cfg    Config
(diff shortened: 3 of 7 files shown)


pkg/medialib/radarr/client.go
```diff
@@ -0,0 +1,81 @@
...
+
+// New creates a new Radarr client.
+func New(cfg Config) *Client {
+	s := starr.New(cfg.APIKey, cfg.URL, 0)
+	return &Client{cfg: cfg, radarr: radarrlib.New(s)}
+}
+
+// GetMovieByFilePath returns the movie whose file matches path.
+// Returns medialib.ErrNotFound if no match is found.
+func (c *Client) GetMovieByFilePath(ctx context.Context, path string) (medialib.Movie, error) {
```
solidDoWant, line 48: Here and elsewhere in this PR, is there a path transversal risk? The os.Root type could be useful for mitigating this.

pkg/medialib/medialib.go
```diff
@@ -1,2 +1,37 @@
...
+	"context"
+	"errors"
+)
+
+// ErrNotFound is returned when a media item is not found in the library.
+var ErrNotFound = errors.New("not found in library")
+
+// Movie represents a movie entry in a movie library service.
+type Movie struct {
+	ID    int64
```
solidDoWant, line 15: What "ID" is this? Needs to be clear to developers. See episode struct as well

pkg/medialib/radarr/client_test.go
```diff
@@ -0,0 +1,168 @@
...
+	t.Cleanup(srv.Close)
+
+	client := radarr.New(radarr.Config{URL: srv.URL, APIKey: "test-key"})
+
+	err := client.RefreshMovie(t.Context(), 42)
+	require.NoError(t, err)
+}
+
+func TestGetMovieByFilePath_UnreachableURL(t *testing.T) {
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {}))
```
solidDoWant: Is it even necessary to start a server for this test?

pkg/medialib/sonarr/client.go
```diff
@@ -0,0 +1,101 @@
...
+		return medialib.Episode{}, fmt.Errorf("list series: %w", err)
+	}
+
+	for _, s := range series {
+		files, err := c.sonarr.GetSeriesEpisodeFilesContext(ctx, s.ID)
+		if err != nil {
+			return medialib.Episode{}, fmt.Errorf("list episode files for series %d: %w", s.ID, err)
+		}
+
+		for _, f := range files {
```
solidDoWant: f -> file

pkg/medialib/sonarr/client.go
```diff
@@ -0,0 +1,101 @@
...
+			path = c.cfg.RemotePathPrefix + after
+		}
+	}
+
+	series, err := c.sonarr.GetAllSeriesContext(ctx)
+	if err != nil {
+		return medialib.Episode{}, fmt.Errorf("list series: %w", err)
+	}
+
+	for _, s := range series {
```
solidDoWant: This could potentially iterate over a lot of files and a lot of API calls. Do you have any suggestions on how this could be reduced?

go.mod
```diff
@@ -76,6 +76,7 @@ require (
 	golang.org/x/sync v0.20.0 // indirect
 	golang.org/x/sys v0.42.0 // indirect
 	golang.org/x/text v0.35.0 // indirect
+	golift.io/starr v1.3.1 // indirect
```
solidDoWant: you probably need to run `go mod tidy`

pkg/medialib/radarr/client.go
solidDoWant: No file I/O is performed here — the path is only compared as a string against Radarr's stored paths, so there is no filesystem path traversal risk. `os.Root` mitigates attacks during I/O operations, which don't apply here. That said, I've added `filepath.Clean` after translation to normalize any `.`/`..` components, which keeps comparisons consistent regardless of how the caller constructs the path.

pkg/medialib/medialib.go
solidDoWant: Added doc comments: `Movie.ID` is now documented as "the internal database ID assigned by the backing movie library service", and similarly for `Episode.ID`.

pkg/medialib/radarr/client_test.go
solidDoWant: Good call — simplified to grab a free port via `net.Listen`, close the listener, then use that URL. No server is started.

pkg/medialib/sonarr/client.go
solidDoWant: Renamed.

pkg/medialib/sonarr/client.go
solidDoWant: The Sonarr `Series` type has a `Path` field that stores the series root directory (e.g. `/tv/Breaking Bad`). I now use that as a prefix filter: after fetching all series (1 API call), the implementation checks `strings.HasPrefix(filePath, seriesPath+"/")` and only fetches episode files for the one matching series (1 call). This reduces total API calls from O(N series) to 3 in the common case (series list + episode files for one series + episode lookup).

go.mod
solidDoWant: Done — ran `go mod tidy` which moves `golift.io/starr` from `// indirect` to the direct `require` block.

pkg/medialib/radarr/client.go
solidDoWant: What about if Radar has a filepath transversal vulnerability? Should the program verify that the cleaned path is under `RemotePathPrefix`, or is this unnecessary?

pkg/medialib/sonarr/client.go
solidDoWant: This program will eventually be used during the download process, which will look something like this:
1. User requests media via Radarr
2. Radarr searches for the media and informs the download client of what to download
3. Download client downloads the media
4. This program detects the downloaded file, performs operations on it, and moves it to a different directory
5. Radarr sees the file in the new directory, and handles importing it

In this case, Radarr won't have imported the file by the time all these API calls are made (here and for movies). Will this approach still work?

pkg/medialib/radarr/client_test.go
solidDoWant: Is a listener even needed? Could the URL just point to an invalid location?

pkg/medialib/sonarr/client.go
solidDoWant: In the described pre-import workflow, `GetMovieByFilePath` will return `ErrNotFound` for step 4 because Radarr won't have a `MovieFile` record until after step 5 (import). The current implementation only works for files Radarr has already imported into its library.

If the primary use case is pre-import (process → then Radarr imports), we'd need a different approach — e.g. querying Radarr's download queue or matching by TMDB/IMDB metadata from the filename rather than by file path.

Could you clarify the intended call site? Specifically: is `GetMovieByFilePath` expected to be called before or after Radarr imports the file?

pkg/medialib/radarr/client.go
solidDoWant: Good point — added a guard after `filepath.Clean`: when `RemotePathPrefix` is set, the cleaned path must still start with that prefix or we return an error immediately (distinct from `ErrNotFound`). This also covers the case where a malicious Radarr response could craft a path that escapes the intended root. Test case added.

pkg/medialib/radarr/client_test.go
solidDoWant: Simplified — replaced the `net.Listen` helper with a hardcoded `http://127.0.0.1:1` constant. Port 1 requires root privileges to bind and has no listener in any normal environment, so the connection is refused immediately without any setup.

pkg/medialib/sonarr/client.go
solidDoWant: Before. Is this doable via radarr/sonarr, or is an external API needed?

solidDoWant: Yes, doable via Radarr/Sonarr. Both have a `GET /api/v3/parse?path=...` endpoint that identifies a movie/episode from its filename — this is specifically designed for pre-import use and requires no library record for the file.

Sonarr's parse endpoint is already implemented in golift (`ParseContext`). Radarr's is not in the library, but since `radarr.Radarr` embeds `starr.APIer` (which exposes `GetInto`), I can call it directly.

This also simplifies the implementation significantly: 1 API call instead of iterating all movies/series. Rewriting both clients now.

PR #22, opened 2026-03-22
feat(watcher): implement directory scan worker with Hatchet cron dispatch

### LINKED ISSUE #7: feat(watcher): directory watcher with workflow dispatch
## Summary

Implement the full `cmd/watcher` binary: use `fsnotify` to watch configured directories for new files and subdirectories, then submit the corresponding Hatchet workflow run using the file path as the idempotency key.

## Acceptance Criteria

- [x] **Given** a directory is configured with a workflow mapping, **when** a new file appears in that directory, **then** the watcher submits a Hatchet workflow run for the configured workflow within 5 seconds.
- [x] **Given** the same file path appears again (e.g. watcher restart), **when** the workflow is submitted again, **then** Hatchet deduplicates it and no second run is started (idempotency key = file path).
- [x] **Given** a file appears in a directory that has no configured mapping, **when** it is detected, **then** the watcher logs a warning and takes no other action.
- [x] **Given** multiple watcher instances running against the same directories, **when** both detect the same new file, **then** only one workflow run is created (enforced by the idempotency key).
- [x] **Given** the watcher is running, **when** it receives SIGTERM or SIGINT, **then** it shuts down cleanly within 5 seconds.
- [x] **Given** a configured watch directory does not exist at startup, **when** the watcher starts, **then** it exits non-zero with a clear error message.

## Technical Context

- Depends on: #2
- fsnotify library: `github.com/fsnotify/fsnotify`
- Watch for `fsnotify.Create` and `fsnotify.Rename` events (renames are how many tools atomically place files)
- Idempotency key: absolute file path
- Workflow input passed to Hatchet: `{ "file_path": "/watch/movies/example.mkv" }`
- The watcher should watch recursively if a subdirectory is created (some ingestion tools create a folder per title)

## Non-Goals

- No file stability detection (upstream tooling guarantees atomic writes per the spec)
- No polling fallback — fsnotify events only

solidDoWant:
## Summary

- Implements the `cmd/watcher` binary's core functionality: a Hatchet-managed cron workflow that recursively scans configured watch directories and spawns child workflow runs for each file found
- Replaces fsnotify (inotify-based, incompatible with NFS mounts) with a polling approach managed by Hatchet's cron scheduler (6-field cron, seconds granularity)
- Uses `hatchet.WithRunKey(absFilePath)` inside a Hatchet task context for proper server-side idempotency per file path
- Adds configurable `cron_schedule` field to YAML config (default: `*/5 * * * * *`)

## Test plan

- [x] `make test` passes for `cmd/watcher` (all new tests green)
- [x] `make build` produces valid `bin/watcher` binary
- [x] `make lint` reports no issues
- [ ] Integration: start watcher with a valid config and NFS-mounted volume; confirm workflow runs appear in Hatchet UI when files are dropped

Fixes #7

### DIFF
--- a/cmd/watcher/watcher.go
+++ b/cmd/watcher/watcher.go
@@ -50,13 +50,20 @@ func NewScanWorkflow(client *hatchet.Client, cfg *Config) *hatchet.StandaloneTas
 }
 
 // validateWatchDirs returns an error listing every configured watch directory that does
-// not exist, so the operator sees all problems at once rather than fixing them one at a time.
+// not exist or is not a directory, so the operator sees all problems at once rather than
+// fixing them one at a time.
 func validateWatchDirs(cfg *Config) error {
 	errs := make([]error, 0, len(cfg.Watches))
 
 	for _, w := range cfg.Watches {
-		if _, err := os.Stat(w.Path); err != nil {
+		info, err := os.Stat(w.Path)
+		if err != nil {
 			errs = append(errs, fmt.Errorf("watch directory %q: %w", w.Path, err))
+			continue
+		}
+
+		if !info.IsDir() {
+			errs = append(errs, fmt.Errorf("watch path %q is not a directory", w.Path))
 		}
 	}
 
@@ -66,12 +73,21 @@ func validateWatchDirs(cfg *Config) error {
 // scan walks every configured watch directory recursively and calls dispatch for each
 // file found. Files in subdirectories inherit the mapping of their watch root. All
 // per-file errors (access errors and dispatch errors) are collected and returned as an
-// aggregate so a single failure does not abort the scan.
+// aggregate so a single failure does not abort the scan. The walk respects ctx
+// cancellation and returns immediately on shutdown.
 func scan(ctx context.Context, cfg *Config, dispatch dispatchFunc) error {
 	var errs []error
 
 	for _, w := range cfg.Watches {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+
 		if err := filepath.WalkDir(w.Path, func(path string, d fs.DirEntry, err error) error {
+			if ctxErr := ctx.Err(); ctxErr != nil {
+				return ctxErr
+			}
+
 			if err != nil {
 				errs = append(errs, fmt.Errorf("scan error at %q: %w", path, err))
 				return nil
@@ -93,6 +109,10 @@ func scan(ctx context.Context, cfg *Config, dispatch dispatchFunc) error {
 
 			return nil
 		}); err != nil {
+			if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+				return err
+			}
+
 			errs = append(errs, fmt.Errorf("walk directory %q: %w", w.Path, err))
 		}
 	}
--- a/cmd/watcher/watcher.go
+++ b/cmd/watcher/watcher.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io/fs"
 	"log"
@@ -22,7 +23,7 @@ type dispatchFunc func(ctx context.Context, workflowName, filePath string) error
 // Overlapping scans are dropped (CANCEL_NEWEST, max 1 concurrent run) so a slow
 // scan does not pile up behind a cron backlog.
 func NewScanWorkflow(client *hatchet.Client, cfg *Config) *hatchet.StandaloneTask {
-	var maxRuns int32 = 1
+	maxRuns := int32(1)
 	strategy := types.CancelNewest
 
 	return client.NewStandaloneTask(
@@ -49,14 +50,18 @@ func NewScanWorkflow(client *hatchet.Client, cfg *Config) *hatchet.StandaloneTas
 	)
 }
 
-// validateWatchDirs returns an error if any configured watch directory does not exist.
+// validateWatchDirs returns an error listing every configured watch directory that does
+// not exist, so the operator sees all problems at once rather than fixing them one at a time.
 func validateWatchDirs(cfg *Config) error {
+	var errs []error
+
 	for _, w := range cfg.Watches {
 		if _, err := os.Stat(w.Path); err != nil {
-			return fmt.Errorf("watch directory %q: %w", w.Path, err)
+			errs = append(errs, fmt.Errorf("watch directory %q: %w", w.Path, err))
 		}
 	}
-	return nil
+
+	return errors.Join(errs...)
 }
 
 // scan walks every configured watch directory recursively and calls dispatch for each
@@ -69,16 +74,20 @@ func scan(ctx context.Context, cfg *Config, dispatch dispatchFunc) error {
 				log.Printf("warning: scan error at %q: %v", path, err)
 				return nil
 			}
+
 			if d.IsDir() {
 				return nil
 			}
+
 			absPath, err := filepath.Abs(path)
 			if err != nil {
 				return fmt.Errorf("resolve absolute path for %q: %w", path, err)
 			}
+
 			if dispatchErr := dispatch(ctx, w.Workflow, absPath); dispatchErr != nil {
 				log.Printf("warning: dispatch workflow %q for %q: %v", w.Workflow, absPath, dispatchErr)
 			}
+
 			return nil
 		}); err != nil {
 			return fmt.Errorf("walk directory %q: %w", w.Path, err)
--- a/cmd/watcher/watcher.go
+++ b/cmd/watcher/watcher.go
@@ -97,5 +97,9 @@ func scan(ctx context.Context, cfg *Config, dispatch dispatchFunc) error {
 		}
 	}
 
-	return errors.Join(errs...)
+	if err := errors.Join(errs...); err != nil {
+		return fmt.Errorf("directory scan completed with errors: %w", err)
+	}
+
+	return nil
 }
--- a/cmd/watcher/watcher.go
+++ b/cmd/watcher/watcher.go
@@ -1,0 +1,88 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"io/fs"
+	"log"
+	"os"
+	"path/filepath"
+
+	"github.com/hatchet-dev/hatchet/pkg/client/types"
+	hatchet "github.com/hatchet-dev/hatchet/sdks/go"
+)
+
+// dispatchFunc submits a workflow run for the given absolute file path.
+type dispatchFunc func(ctx context.Context, workflowName, filePath string) error
+
+// NewScanWorkflow returns a Hatchet standalone task that scans all configured watch
+// directories on the configured cron schedule and spawns a child workflow run for
+// every file found, using the absolute file path as the idempotency key.
+//
+// Overlapping scans are dropped (CANCEL_NEWEST, max 1 concurrent run) so a slow
+// scan does not pile up behind a cron backlog.
+func NewScanWorkflow(client *hatchet.Client, cfg *Config) *hatchet.StandaloneTask {
+	var maxRuns int32 = 1
+	strategy := types.CancelNewest
+
+	return client.NewStandaloneTask(
+		"directory-scan",
+		func(ctx hatchet.Context, _ struct{}) (struct{}, error) {
+			dispatch := func(dispatchCtx context.Context, workflowName, filePath string) error {
+				_, err := client.RunNoWait(
+					dispatchCtx,
+					workflowName,
+					map[string]string{"file_path": filePath},
+					hatchet.WithRunKey(filePath),
+				)
+				return err
+			}
+			return struct{}{}, scan(ctx, cfg, dispatch)
+		},
+		hatchet.WithWorkflowCron(cfg.CronSchedule),
+		hatchet.WithWorkflowConcurrency(types.Concurrency{
+			// Constant expression groups all scan runs under a single concurrency slot.
+			Expression:    `"scan"`,
+			MaxRuns:       &maxRuns,
+			LimitStrategy: &strategy,
+		}),
+	)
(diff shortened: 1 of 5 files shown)


cmd/watcher/watcher.go
```diff
@@ -0,0 +1,88 @@
...
+type dispatchFunc func(ctx context.Context, workflowName, filePath string) error
+
+// NewScanWorkflow returns a Hatchet standalone task that scans all configured watch
+// directories on the configured cron schedule and spawns a child workflow run for
+// every file found, using the absolute file path as the idempotency key.
+//
+// Overlapping scans are dropped (CANCEL_NEWEST, max 1 concurrent run) so a slow
+// scan does not pile up behind a cron backlog.
+func NewScanWorkflow(client *hatchet.Client, cfg *Config) *hatchet.StandaloneTask {
+	var maxRuns int32 = 1
```
solidDoWant: Why not just `maxRuns := 1`?

cmd/watcher/watcher.go
```diff
@@ -0,0 +1,88 @@
...
+			LimitStrategy: &strategy,
+		}),
+	)
+}
+
+// validateWatchDirs returns an error if any configured watch directory does not exist.
+func validateWatchDirs(cfg *Config) error {
+	for _, w := range cfg.Watches {
+		if _, err := os.Stat(w.Path); err != nil {
+			return fmt.Errorf("watch directory %q: %w", w.Path, err)
```
solidDoWant: Should this collect all errors for all dirs or just return the first one?

cmd/watcher/watcher.go
```diff
@@ -0,0 +1,88 @@
...
+			}
+			if dispatchErr := dispatch(ctx, w.Workflow, absPath); dispatchErr != nil {
+				log.Printf("warning: dispatch workflow %q for %q: %v", w.Workflow, absPath, dispatchErr)
+			}
+			return nil
+		}); err != nil {
+			return fmt.Errorf("walk directory %q: %w", w.Path, err)
+		}
+	}
+	return nil
```
solidDoWant, line 124: space out these statements more. Is the a linter rule that can be configured for this?

cmd/watcher/watcher.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// scan walks every configured watch directory recursively and calls dispatch for each
+// file found. Files in subdirectories inherit the mapping of their watch root. Dispatch
+// errors are logged as warnings so a single failed submission does not abort the scan.
+func scan(ctx context.Context, cfg *Config, dispatch dispatchFunc) error {
+	for _, w := range cfg.Watches {
+		if err := filepath.WalkDir(w.Path, func(path string, d fs.DirEntry, err error) error {
```
solidDoWant, line 86: Do you have an opinion on using `os.Root` for any of these operations?

cmd/watcher/watcher.go
solidDoWant: Changed to `maxRuns := int32(1)`. The field type is `*int32` so we need an `int32` variable to take its address — `maxRuns := 1` would infer `int`, giving a `*int` which doesn't satisfy `*int32`.

cmd/watcher/watcher.go
solidDoWant: Changed to collect all errors before returning, using `errors.Join`. A new test case verifies that both missing-directory paths appear in the combined error message.

cmd/watcher/watcher.go
solidDoWant: Added blank lines between each logical section in the callback. For automatic enforcement, golangci-lint's `wsl` linter can require blank lines before certain statements — but it's quite opinionated and would likely require widespread changes across the codebase. I'd recommend enabling it separately if you want it rather than bundling it here.

cmd/watcher/watcher.go
solidDoWant: My opinion: not necessary here. `os.Root` (Go 1.24+) would sandbox filesystem ops to prevent symlinks from escaping the watch directory — useful defense-in-depth. But the practical risk is low: the watch dirs are trusted config-controlled NFS mounts, and even if a symlink pointed outside the root, the worst outcome is that a workflow run is dispatched for an unexpected path (which would simply fail to transcode).

The implementation cost is non-trivial: `os.Root` doesn't expose a Walk method, so you'd need to implement recursive directory reading manually with `root.ReadDir` + recursion to replace `filepath.WalkDir`.

I'd leave it as-is for now and revisit if there's ever a concrete threat model that warrants the extra complexity. Happy to implement it if you disagree.

cmd/watcher/watcher.go
```diff
@@ -0,0 +1,97 @@
...
+			MaxRuns:       &maxRuns,
+			LimitStrategy: &strategy,
+		}),
+	)
+}
+
+// validateWatchDirs returns an error listing every configured watch directory that does
+// not exist, so the operator sees all problems at once rather than fixing them one at a time.
+func validateWatchDirs(cfg *Config) error {
+	var errs []error
```
solidDoWant: init to `len(cfg.Watches)`: `errs := make([]error, 0, len(cfg.Watches))`

cmd/watcher/watcher.go
```diff
@@ -0,0 +1,97 @@
...
+// scan walks every configured watch directory recursively and calls dispatch for each
+// file found. Files in subdirectories inherit the mapping of their watch root. Dispatch
+// errors are logged as warnings so a single failed submission does not abort the scan.
+func scan(ctx context.Context, cfg *Config, dispatch dispatchFunc) error {
+	for _, w := range cfg.Watches {
+		if err := filepath.WalkDir(w.Path, func(path string, d fs.DirEntry, err error) error {
+			if err != nil {
+				log.Printf("warning: scan error at %q: %v", path, err)
+				return nil
+			}
```
solidDoWant, line 94: Is this the correct way to handle these errors, where the error is passed via the walk callback?

cmd/watcher/watcher.go
```diff
@@ -0,0 +1,97 @@
...
+
+			absPath, err := filepath.Abs(path)
+			if err != nil {
+				return fmt.Errorf("resolve absolute path for %q: %w", path, err)
+			}
+
+			if dispatchErr := dispatch(ctx, w.Workflow, absPath); dispatchErr != nil {
+				log.Printf("warning: dispatch workflow %q for %q: %v", w.Workflow, absPath, dispatchErr)
+			}
+
```
solidDoWant, line 109: Do you have an opinion on aggregating all errors from this callback function so that `scan` returns an aggregate error with all of them?

cmd/watcher/watcher.go
solidDoWant: Pre-allocated with `make([]error, 0, len(cfg.Watches))` in 3a0eea2.

cmd/watcher/watcher.go
solidDoWant: Yes — returning `nil` from the callback is correct here. Returning a non-nil error would abort the entire `WalkDir`, which is wrong for resilience on NFS mounts. The access error is now appended to the aggregate slice so it surfaces to the caller (3a0eea2).

cmd/watcher/watcher.go
solidDoWant: Done in 3a0eea2. `scan` now collects all per-file errors (access errors from the WalkDir callback, path resolution errors, and dispatch errors) into a local slice and returns `errors.Join` at the end. The walk still processes every file — no early abort. Updated `TestScan_DispatchErrorsAreAggregated` to assert an error is returned while all files are still visited.

cmd/watcher/watcher.go
```diff
@@ -0,0 +1,101 @@
...
+					dispatchCtx,
+					workflowName,
+					map[string]string{"file_path": filePath},
+					hatchet.WithRunKey(filePath),
+				)
+				return err
+			}
+			return struct{}{}, scan(ctx, cfg, dispatch)
+		},
+		hatchet.WithWorkflowCron(cfg.CronSchedule),
```
solidDoWant, line 42: Given that this workflow will run on a (short) schedule, should retries be enabled or disabled?

cmd/watcher/watcher.go
```diff
@@ -0,0 +1,101 @@
...
+				errs = append(errs, fmt.Errorf("dispatch workflow %q for %q: %w", w.Workflow, absPath, dispatchErr))
+			}
+
+			return nil
+		}); err != nil {
+			errs = append(errs, fmt.Errorf("walk directory %q: %w", w.Path, err))
+		}
+	}
+
+	return errors.Join(errs...)
```
solidDoWant: Should this be wrapped with a more descriptive message?

cmd/watcher/watcher.go
solidDoWant: Retries should be **disabled**. Since the scan runs on a short cron schedule, a failed scan will be retried automatically by the next cron trigger — explicit retries would only risk duplicate dispatches before the next tick. The current code doesn't call `WithRetries`, so `config.retries` defaults to `0`. The SDK's internal `declaration.go` guards with `if opts.Retries \!= 0` before setting the retries pointer, so `nil` is sent to the server — effectively no retry configuration. There's currently no way to distinguish "0 retries" from "use server default" via this API; if the server default is non-zero, explicit `WithRetries(0)` would have no effect due to that guard. I've left the code as-is — it already sends no retries config.

cmd/watcher/watcher.go
solidDoWant: Done in 57fab5d. The return now wraps the joined errors: `fmt.Errorf("directory scan completed with errors: %w", errors.Join(errs...))`. This adds top-level context while preserving full `errors.Is`/`errors.As` unwrapping of all individual errors.

cmd/watcher/watcher.go
```diff
@@ -0,0 +1,105 @@
...
+
+			absPath, err := filepath.Abs(path)
+			if err != nil {
+				errs = append(errs, fmt.Errorf("resolve absolute path for %q: %w", path, err))
+				return nil
+			}
+
+			if dispatchErr := dispatch(ctx, w.Workflow, absPath); dispatchErr != nil {
+				errs = append(errs, fmt.Errorf("dispatch workflow %q for %q: %w", w.Workflow, absPath, dispatchErr))
+			}
```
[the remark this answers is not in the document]
solidDoWant: This is out of scope for the MVP. The watched directories should typically be small.

cmd/watcher/watcher.go
```diff
@@ -0,0 +1,105 @@
...
+
+// validateWatchDirs returns an error listing every configured watch directory that does
+// not exist, so the operator sees all problems at once rather than fixing them one at a time.
+func validateWatchDirs(cfg *Config) error {
+	errs := make([]error, 0, len(cfg.Watches))
+
+	for _, w := range cfg.Watches {
+		if _, err := os.Stat(w.Path); err != nil {
+			errs = append(errs, fmt.Errorf("watch directory %q: %w", w.Path, err))
+		}
```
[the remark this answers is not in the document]
solidDoWant: Implemented in a2f8c10. `validateWatchDirs` now calls `os.Stat` and checks `info.IsDir()`, returning `"watch path %q is not a directory"` when the path exists but isn't a directory. The `continue` after the stat error avoids a nil-pointer on `info.IsDir()` for missing paths.

cmd/watcher/watcher_test.go
```diff
@@ -0,0 +1,178 @@
...
+		},
+		{
+			name: "missing directory returns error",
+			cfg: &Config{
+				Watches: []WatchEntry{
+					{Path: "/nonexistent/path/abc123", Workflow: "W"},
+				},
+			},
+			errFunc: require.Error,
+		},
```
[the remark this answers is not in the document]
solidDoWant: Added in a2f8c10: a new table-driven case `"path that exists but is a file returns error"` creates a temp file and asserts the error contains `"not a directory"`.

cmd/watcher/watcher.go
```diff
@@ -0,0 +1,105 @@
...
+	for _, w := range cfg.Watches {
+		if err := filepath.WalkDir(w.Path, func(path string, d fs.DirEntry, err error) error {
+			if err != nil {
+				errs = append(errs, fmt.Errorf("scan error at %q: %w", path, err))
+				return nil
+			}
+
+			if d.IsDir() {
+				return nil
+			}
```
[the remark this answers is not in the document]
solidDoWant: Implemented in a2f8c10. `scan` now checks `ctx.Err()` before starting each watch root and at the top of the WalkDir callback. When cancelled, the callback returns the context error, which stops the walk immediately; the outer handler then checks `errors.Is(err, context.Canceled)` (and `DeadlineExceeded`) and returns the error directly rather than adding it to the aggregate slice. Added `TestScan_ContextCancellationStopsWalk` to verify the returned error satisfies `errors.Is(err, context.Canceled)`.

cmd/watcher/watcher.go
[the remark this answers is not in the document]
solidDoWant: Valid concern for very large trees, but I'd recommend deferring this to a follow-up issue for a few reasons:\n\n1. **Self-deduplicating by design**: workflows move the source file as their final step, so a file that was successfully processed won't be present in the next scan. Steady-state load is bounded by the number of unprocessed files, not the total directory size.\n2. **Approved scope**: the issue and accepted plan explicitly rely on Hatchet's server-side run-key idempotency for dedup. In-memory state or a persisted cursor is a separate design decision.\n3. **In-memory state has its own risks**: state is lost on restart, so a crash during processing could leave files stranded until the next restart flushes the set.\n\nIf the directory size becomes a real operational concern (e.g., large backlog of unprocessed files), an incremental scan based on file mtime or a bloom filter would be the right approach — but that deserves its own issue with proper acceptance criteria.

PR #24, opened 2026-03-25
feat(workflow): show/episode processing workflow

### LINKED ISSUE #9: feat(workflow): show/episode processing workflow
## Summary

Implement `ShowWorkflow` in `workflows/show.go` — the Hatchet workflow for ingested TV episode files. Mirrors `MovieWorkflow` but uses the Sonarr API instead of Radarr.

## Acceptance Criteria

- [ ] **Given** a file path to a valid episode file that Sonarr knows about, **when** the workflow runs to completion, **then** a transcoded MKV file exists in the configured output directory with the original filename stem and a `.mkv` extension.
- [ ] **Given** the input file is already H.264 or H.265 in an MKV container, **when** the workflow runs, **then** the video stream is copied without re-encoding.
- [ ] **Given** the input file is not H.264 or H.265, **when** the workflow runs, **then** the video stream is transcoded to H.265.
- [ ] **Given** the workflow completes successfully, **when** inspecting the source directory, **then** the original source file no longer exists.
- [ ] **Given** the workflow completes successfully, **when** Sonarr is checked, **then** `RefreshSeries` has been called and Sonarr reflects the new file.
- [ ] **Given** Sonarr does not know about the file (returns `ErrNotFound`), **when** the workflow runs, **then** it fails at the lookup step and invokes the failure webhook.
- [ ] **Given** any step fails, **when** the workflow terminates, **then** the failure webhook is called with the step name and error, and the source file is not deleted.
- [ ] **Given** a file path that exists but is not a valid video file (e.g. a text file), **when** the workflow runs, **then** the file is deleted and no other action is taken (no failure webhook, no downstream steps).
- [ ] **Given** the workflow is submitted with a file path that is already being processed (duplicate submission), **when** Hatchet receives it, **then** only one run executes (idempotency enforced by #7).

## Technical Context

- Depends on: #2, #3, #4, #5, #6
- Output directory sourced from env var: `SHOW_OUTPUT_DIR`
- Workflow steps (in order):
  1. `probe` — `pkg/ffprobe`; on ffprobe error or no video streams: delete file, return `is_valid_media=false`, skip all downstream steps without triggering `OnFailure`; propagate `context.Canceled`/`DeadlineExceeded` as real errors instead
  2. `lookup` — `pkg/medialib` Sonarr client; fail with `ErrNotFound` if not in library; returns `series_id` for use by `notify`
  3. `transcode` — same H.264/H.265 copy-or-transcode logic as `MovieWorkflow`; output filename is always `<stem>.mkv` regardless of input extension; write via temp file (`._<stem>.mkv.tmp`) then atomic rename
  4. `notify` — `sonarr.RefreshSeries(series_id)`
  5. `cleanup` — delete source file
- Concurrency key on `input.file_path` with `CancelNewest`, `maxRuns=1` (idempotency)
- `OnFailure`: sort step names before joining into the webhook payload (deterministic `Step` field)
- Hatchet workflow name: `"ShowWorkflow"`

## Non-Goals

- No multi-episode file handling
- No subtitle extraction or generation

solidDoWant:
## Summary
- Extract shared probe/transcode/cleanup step logic from `workflows/movie` into new `workflows/shared` package (per review feedback)
- Refactor `workflows/movie` to delegate to `workflows/shared` — no behaviour change
- Add `workflows/show`: `ShowWorkflow` mirrors `MovieWorkflow` but uses the Sonarr `EpisodeLibrary` (`GetEpisodeByFilePath` + `RefreshSeries`), with concurrency key on `file_path` (`CancelNewest`, `maxRuns=1`) and sorted step names in `OnFailure`

## Test plan
- [x] Unit tests pass: `workflows/shared` (probe, transcode, cleanup, selectVideoCodec, NotifyWorkflowFailure), `workflows/show` (lookup), `workflows/movie` (lookup)
- [x] Integration tests in `workflows/show/integration_test.go` cover: valid video transcoded + source deleted, `RefreshSeries` called, non-video file deleted (no webhook/downstream), lookup failure causes workflow failure + source preserved
- [x] `make build`, `make lint` clean; `make test` clean except pre-existing `pkg/ffprobe/TestProbe_CancelledContext` failure unrelated to this PR

Fixes #9

### DIFF
--- a/workflows/shared/cleanup.go
+++ b/workflows/shared/cleanup.go
@@ -1,4 +1,4 @@
-package movie
+package shared
 
 import (
 	"errors"
@@ -6,8 +6,8 @@ import (
 	"os"
 )
 
-// runCleanup deletes the original source file after successful processing.
-func runCleanup(filePath string) error {
+// RunCleanup deletes the original source file after successful processing.
+func RunCleanup(filePath string) error {
 	if err := os.Remove(filePath); err != nil && !errors.Is(err, os.ErrNotExist) {
 		return fmt.Errorf("delete source file: %w", err)
 	}
--- a/workflows/shared/onfailure.go
+++ b/workflows/shared/onfailure.go
@@ -1,0 +1,45 @@
+package shared
+
+import (
+	"errors"
+	"fmt"
+	"sort"
+	"strings"
+
+	hatchet "github.com/hatchet-dev/hatchet/sdks/go"
+
+	"github.com/solidDoWant/media-processor/pkg/webhook"
+)
+
+// NotifyWorkflowFailure sends an aggregated failure notification to webhookClient.
+// It is intended to be called from a workflow's OnFailure handler.
+// workflowName is included in the webhook payload; filePath is the file being processed.
+// Returns nil (no-op) when there are no step errors.
+func NotifyWorkflowFailure(ctx hatchet.Context, workflowName, filePath string, webhookClient *webhook.Client) error {
+	stepErrors := ctx.StepRunErrors()
+	if len(stepErrors) == 0 {
+		return nil
+	}
+
+	steps := make([]string, 0, len(stepErrors))
+	for stepName := range stepErrors {
+		steps = append(steps, stepName)
+	}
+	sort.Strings(steps)
+
+	errs := make([]error, 0, len(stepErrors))
+	for _, stepName := range steps {
+		errs = append(errs, fmt.Errorf("%s: %s", stepName, stepErrors[stepName]))
+	}
+
+	if err := webhookClient.NotifyFailure(ctx, webhook.FailureEvent{
+		Workflow: workflowName,
+		FilePath: filePath,
+		Step:     strings.Join(steps, ", "),
+		Err:      errors.Join(errs...),
+	}); err != nil {
+		return fmt.Errorf("notify failure: %w", err)
+	}
+
+	return nil
+}
--- a/workflows/shared/onfailure.go
+++ b/workflows/shared/onfailure.go
@@ -1,22 +1,20 @@
 package shared
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"sort"
 	"strings"
 
-	hatchet "github.com/hatchet-dev/hatchet/sdks/go"
-
 	"github.com/solidDoWant/media-processor/pkg/webhook"
 )
 
 // NotifyWorkflowFailure sends an aggregated failure notification to webhookClient.
-// It is intended to be called from a workflow's OnFailure handler.
+// stepErrors is the map returned by ctx.StepRunErrors() in an OnFailure handler.
 // workflowName is included in the webhook payload; filePath is the file being processed.
-// Returns nil (no-op) when there are no step errors.
-func NotifyWorkflowFailure(ctx hatchet.Context, workflowName, filePath string, webhookClient *webhook.Client) error {
-	stepErrors := ctx.StepRunErrors()
+// Returns nil (no-op) when stepErrors is empty.
+func NotifyWorkflowFailure(ctx context.Context, stepErrors map[string]string, workflowName, filePath string, webhookClient *webhook.Client) error {
 	if len(stepErrors) == 0 {
 		return nil
 	}
--- a/workflows/shared/testhelpers_test.go
+++ b/workflows/shared/testhelpers_test.go
@@ -1,0 +1,24 @@
+package shared
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// testVideoPath points to the small H.264/MP4 clip shared with the ffprobe package.
+const testVideoPath = "../../pkg/ffprobe/testdata/video.mp4"
+
+// copyTestVideo copies the shared test video to a temp file and returns its path.
+// The file is NOT registered for cleanup so tests can verify deletion behaviour.
+func copyTestVideo(t *testing.T) string {
+	t.Helper()
+	src, err := os.ReadFile(testVideoPath)
+	require.NoError(t, err)
+
+	dst := filepath.Join(t.TempDir(), "video.mp4")
+	require.NoError(t, os.WriteFile(dst, src, 0o600))
+	return dst
+}
--- a/workflows/shared/transcode.go
+++ b/workflows/shared/transcode.go
@@ -29,8 +29,10 @@ func SelectVideoCodec(videoCodecName, format string) ffmpeg.Codec {
 // The output always carries a .mkv extension to match the forced MKV container.
 // Writing directly to the output directory avoids a cross-filesystem copy and
 // guarantees the rename is atomic on Linux (same directory).
-func RunTranscode(ctx context.Context, filePath string, probe ProbeOutput, outputDir string) error {
-	videoCodec := SelectVideoCodec(probe.VideoCodec, probe.Format)
+// videoCodecName and format are the codec name and container format from ffprobe
+// (e.g. "h264", "matroska,webm").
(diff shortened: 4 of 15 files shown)


workflows/shared/testhelpers_test.go
```diff
@@ -0,0 +1,24 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+// testVideoPath points to the small H.264/MP4 clip shared with the ffprobe package.
+const testVideoPath = "../../pkg/ffprobe/testdata/video.mp4"
+
+// copyTestVideo copies the shared test video to a temp file and returns its path.
+// The file is NOT registered for cleanup so tests can verify deletion behaviour.
+func copyTestVideo(t *testing.T) string {
```
solidDoWant, line 16: Should this register it for cleanup anyway, so that even if the test fails it is cleaned up?

workflows/shared/cleanup.go
solidDoWant, line 1: Should this package live under `pkg`? Why or why not? Is there value to these resources outside of the project?

workflows/shared/transcode.go
```diff
@@ -24,20 +24,20 @@ func selectVideoCodec(videoCodecName, format string) ffmpeg.Codec {
...
 
-// runTranscode transcodes input.FilePath into outputDir, writing to a temp file named
+// RunTranscode transcodes filePath into outputDir, writing to a temp file named
 // "._<stem>.mkv.tmp" and atomically renaming it to "<stem>.mkv" on success.
 // The output always carries a .mkv extension to match the forced MKV container.
 // Writing directly to the output directory avoids a cross-filesystem copy and
 // guarantees the rename is atomic on Linux (same directory).
-func runTranscode(ctx context.Context, input MovieInput, probe probeOutput, outputDir string) error {
-	videoCodec := selectVideoCodec(probe.VideoCodec, probe.Format)
+func RunTranscode(ctx context.Context, filePath string, probe ProbeOutput, outputDir string) error {
```
solidDoWant: Instead of `ProbeOutput`, should this take specific field(s) to decouple the two?

workflows/show/show.go
```diff
@@ -0,0 +1,140 @@
...
+		if err := webhookClient.NotifyFailure(ctx, webhook.FailureEvent{
+			Workflow: showWorkflowName,
+			FilePath: input.FilePath,
+			Step:     strings.Join(steps, ", "),
+			Err:      errors.Join(errs...),
+		}); err != nil {
+			return struct{}{}, fmt.Errorf("notify failure: %w", err)
+		}
+
+		return struct{}{}, nil
```
solidDoWant: Both this and the movie workflow follow the same logic. Should this function be extracted and shared?

workflows/shared/testhelpers_test.go
solidDoWant: The file lives inside , which the testing framework registers for automatic cleanup via  — so it will be deleted even if the test fails. The comment means we intentionally don't *also* call  on the specific file, because tests that verify deletion behaviour need to check whether the step under test deleted it (not just the deferred cleanup). No additional registration is needed.

solidDoWant: The file lives inside `t.TempDir()`, which the testing framework registers for automatic cleanup — so it will be deleted even if the test fails. The comment means we intentionally do not also separately register the specific file for cleanup, because tests that verify deletion behaviour need to check whether the step under test deleted it (before cleanup fires). No additional registration is needed.

workflows/shared/cleanup.go
solidDoWant: No — `workflows/shared` belongs under `workflows/` because it is workflow-internal plumbing: `ProbeOutput` is a Hatchet step output type (serialized by the orchestrator), `RunProbe` implements the skip-on-invalid-media signal that drives Hatchet’s `WithSkipIf` conditions, and `NotifyWorkflowFailure` calls `hatchet.Context`. These are tightly coupled to the workflow execution model and have no value outside it.

`pkg/` is reserved for standalone, reusable packages without workflow concerns — like `pkg/ffmpeg` and `pkg/ffprobe`, which wrap CLI tools and have stable, workflow-agnostic APIs.

workflows/shared/transcode.go
solidDoWant: Good call — done. `RunTranscode` now accepts `videoCodecName, format string` directly (748ea6e), removing the coupling to `ProbeOutput`.

workflows/show/show.go
solidDoWant: Yes — done. Extracted to `shared.NotifyWorkflowFailure(ctx, workflowName, filePath, webhookClient)` in `workflows/shared/onfailure.go` (748ea6e). Both `movie` and `show` now delegate to it from their `OnFailure` handler.

workflows/shared/onfailure.go
```diff
@@ -0,0 +1,45 @@
...
+	hatchet "github.com/hatchet-dev/hatchet/sdks/go"
+
+	"github.com/solidDoWant/media-processor/pkg/webhook"
+)
+
+// NotifyWorkflowFailure sends an aggregated failure notification to webhookClient.
+// It is intended to be called from a workflow's OnFailure handler.
+// workflowName is included in the webhook payload; filePath is the file being processed.
+// Returns nil (no-op) when there are no step errors.
+func NotifyWorkflowFailure(ctx hatchet.Context, workflowName, filePath string, webhookClient *webhook.Client) error {
```
solidDoWant: This needs unit test(s)

solidDoWant: Done (2a1b31c). `hatchet.Context` has unexported interface methods (`client()`, `action()`) that make it impossible to stub from outside the package, so I refactored `NotifyWorkflowFailure` to accept `(context.Context, stepErrors map[string]string, ...)` instead — the caller passes `ctx.StepRunErrors()` explicitly. This decouples the function from Hatchet entirely and makes it straightforwardly testable.

Added table-driven tests in `onfailure_test.go` covering: no-op on empty/nil errors, correct sorted+joined payload for single and multiple step errors, and webhook error propagation.

workflows/show/integration_test.go
```diff
@@ -0,0 +1,146 @@
...
+	// The probe step marks IsValidMedia=false without error; all downstream steps are skipped.
+	_, err = wf.Run(ctx, ShowInput{FilePath: inputPath})
+	require.NoError(t, err)
+
+	// The probe step deleted the unrecognised file.
+	_, statErr := os.Stat(inputPath)
+	assert.True(t, os.IsNotExist(statErr), "non-video file should be deleted by probe step")
+
+	// Nothing should have been written to the output directory.
+	entries, _ := os.ReadDir(outputDir)
```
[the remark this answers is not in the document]
solidDoWant: Fixed (be1b22a) — `entries, err := os.ReadDir(outputDir)` with `require.NoError(t, err)` before the `assert.Empty` check.

workflows/show/integration_test.go
```diff
@@ -0,0 +1,146 @@
...
+	worker, err := client.NewWorker("test-show-worker", hatchet.WithWorkflows(wf))
+	require.NoError(t, err, "create Hatchet worker")
+
+	cleanup, err := worker.Start()
+	require.NoError(t, err, "start worker")
+	t.Cleanup(func() { _ = cleanup() })
+
+	// Allow time for the worker to register its workflow definitions with the Hatchet server.
+	time.Sleep(3 * time.Second)
+}
```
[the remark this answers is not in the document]
solidDoWant: Done (be1b22a). Replaced the fixed sleep with `require.Eventually` polling `client.Workers().List` until the worker appears with `Status == rest.ACTIVE` (250ms interval, 30s timeout). This makes registration wait as short as needed while bounding the maximum wait on slow CI. All four integration tests still pass.

PR #28, opened 2026-03-26
feat(config): generate JSON schema for watcher config file

### LINKED ISSUE #27: feat(config): generate JSON schema for watcher config file
## Summary

Add a JSON Schema for the watcher YAML config file, automatically derived from the `Config` Go struct using `github.com/invopop/jsonschema`. This allows editor tooling (e.g. VS Code with the RedHat YAML extension) to validate and provide autocompletion for config files. As part of this work, add runtime field-level validation to `loadConfig` so that invalid config (e.g. empty fields, malformed cron expressions) is caught at startup rather than at workflow dispatch time.

## Acceptance Criteria

- [x] **Given** the `Config` and `WatchEntry` structs have `jsonschema` struct tags, **when** `make generate-schema` is run, **then** `schemas/watcher.schema.json` is created/updated with a valid JSON Schema.
- [x] **Given** the generated schema, **when** its property names are inspected, **then** all names match the `yaml` struct tags (e.g. `cron_schedule`, `watches`, `path`, `workflow`).
- [x] **Given** the generated schema, **when** its `required` arrays are inspected, **then** `watches` is required at the top level, and each entry's `path` and `workflow` are required, while `cron_schedule` is optional.
- [x] **Given** the generated schema, **when** `cron_schedule` is present, **then** the schema constrains it to a valid 6-field Hatchet cron expression (seconds-leading format, e.g. `*/5 * * * * *`).
- [x] **Given** the generated schema, **when** `path` or `workflow` fields are inspected, **then** the schema constrains them to non-empty strings.
- [x] **Given** a config file with an empty `path` or `workflow` in a watch entry, **when** `loadConfig` is called, **then** it returns a descriptive error.
- [x] **Given** a config file with a `cron_schedule` value that is not a valid 6-field cron expression, **when** `loadConfig` is called, **then** it returns a descriptive error.
- [x] **Given** the watcher config structs are modified, **when** `make generate-schema` is re-run, **then** the schema reflects those changes.

## Technical Context

- Depends on: none
- Config structs (`Config`, `WatchEntry`) are currently in `package main` (`cmd/watcher/config.go`); they must be moved to an exported internal package (e.g. `internal/watcherconfig`) so a separate generator binary can import them.
- Use `github.com/invopop/jsonschema` with `Reflector{FieldNameTag: "yaml", RequiredFromJSONSchemaTags: true}` — same library and pattern as the `backup-tool` project.
- Add `jsonschema` struct tags to annotate required/optional fields and constraints (e.g. `jsonschema:"minLength=1"` for non-empty strings, a `pattern` for the 6-field cron format).
- Add a `cmd/gen-config-schema/main.go` binary that reflects on `Config` and writes the schema to stdout.
- Makefile target `generate-schema` runs the generator and writes output to `schemas/watcher.schema.json`, pretty-printed via `jq`.
- `schemas/watcher.schema.json` should be committed to the repository.
- For runtime cron validation, a cron parsing library (e.g. `github.com/robfig/cron/v3`) or a regex against the 6-field format is appropriate; choose whichever is lighter.

## Non-Goals

- Schema generation for environment variables or worker configuration.
- Committing VS Code workspace settings or YAML file URI associations.
- Validating that `path` directories exist at schema-generation time (already handled by `validateWatchDirs` at startup).

solidDoWant:
Fixes #27

### DIFF
--- a/cmd/gen-config-schema/main.go
+++ b/cmd/gen-config-schema/main.go
@@ -1,0 +1,28 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"github.com/invopop/jsonschema"
+
+	"github.com/solidDoWant/media-processor/internal/watcherconfig"
+)
+
+func main() {
+	r := &jsonschema.Reflector{
+		FieldNameTag:               "yaml",
+		RequiredFromJSONSchemaTags: true,
+	}
+
+	schema := r.Reflect(&watcherconfig.Config{})
+
+	out, err := json.Marshal(schema)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "failed to marshal schema: %v\n", err)
+		os.Exit(1)
+	}
+
+	fmt.Println(string(out))
+}
--- a/cmd/gen-config-schema/main.go
+++ b/cmd/gen-config-schema/main.go
@@ -12,8 +12,7 @@ import (
 
 func main() {
 	r := &jsonschema.Reflector{
-		FieldNameTag:               "yaml",
-		RequiredFromJSONSchemaTags: true,
+		FieldNameTag: "yaml",
 	}
 
 	schema := r.Reflect(&watcherconfig.Config{})
--- a/cmd/watcher/config.go
+++ b/cmd/watcher/config.go
@@ -3,26 +3,21 @@ package main
 import (
 	"fmt"
 	"os"
+	"regexp"
 
 	"gopkg.in/yaml.v3"
+
+	"github.com/solidDoWant/media-processor/internal/watcherconfig"
 )
 
-// WatchEntry maps a filesystem path to a Hatchet workflow name.
-type WatchEntry struct {
-	Path     string `yaml:"path"`
-	Workflow string `yaml:"workflow"`
-}
+// Type aliases so package-level tests and code continue to reference Config/WatchEntry directly.
+type Config = watcherconfig.Config
+type WatchEntry = watcherconfig.WatchEntry
 
-const defaultCronSchedule = "*/5 * * * * *"
+const defaultCronSchedule = watcherconfig.DefaultCronSchedule
 
-// Config is the top-level watcher configuration loaded from the YAML config file.
-type Config struct {
-	// CronSchedule is the Hatchet cron expression controlling how often the watcher
-	// scans directories (default: every 5 seconds). Supports Hatchet's 6-field format
-	// with a leading seconds field, e.g. "*/5 * * * * *".
-	CronSchedule string       `yaml:"cron_schedule"`
-	Watches      []WatchEntry `yaml:"watches"`
-}
+// sixFieldCron matches a cron expression with exactly 6 space-separated fields.
+var sixFieldCron = regexp.MustCompile(`^(\S+ ){5}\S+$`)
 
 // loadConfig reads and parses the watcher YAML config file at path.
 func loadConfig(path string) (*Config, error) {
@@ -34,6 +29,19 @@ func loadConfig(path string) (*Config, error) {
 	cfg := Config{CronSchedule: defaultCronSchedule}
 	if err := yaml.Unmarshal(data, &cfg); err != nil {
 		return nil, fmt.Errorf("cannot parse config file %q: %w", path, err)
+	}
+
+	for i, w := range cfg.Watches {
+		if w.Path == "" {
+			return nil, fmt.Errorf("watch entry %d: path must not be empty", i)
+		}
+		if w.Workflow == "" {
+			return nil, fmt.Errorf("watch entry %d: workflow must not be empty", i)
+		}
+	}
+
+	if cfg.CronSchedule != "" && !sixFieldCron.MatchString(cfg.CronSchedule) {
+		return nil, fmt.Errorf("cron_schedule %q is not a valid 6-field cron expression (e.g. \"*/5 * * * * *\")", cfg.CronSchedule)
 	}
 
 	return &cfg, nil
--- a/cmd/watcher/config.go
+++ b/cmd/watcher/config.go
@@ -22,12 +22,12 @@ func loadConfig(path string) (*Config, error) {
 	}
 
 	var cfg Config
-	if err := defaults.Set(&cfg); err != nil {
-		return nil, fmt.Errorf("cannot set config defaults: %w", err)
-	}
-
 	if err := yaml.Unmarshal(data, &cfg); err != nil {
 		return nil, fmt.Errorf("cannot parse config file %q: %w", path, err)
+	}
+
+	if err := defaults.Set(&cfg); err != nil {
+		return nil, fmt.Errorf("cannot set config defaults: %w", err)
 	}
 
 	if err := watcherconfig.NewValidator().Struct(cfg); err != nil {
--- a/cmd/watcher/config.go
+++ b/cmd/watcher/config.go
@@ -4,7 +4,7 @@ import (
 	"fmt"
 	"os"
 
-	"github.com/go-playground/validator/v10"
+	"github.com/creasty/defaults"
 	"gopkg.in/yaml.v3"
 
 	"github.com/solidDoWant/media-processor/internal/watcherconfig"
@@ -21,12 +21,16 @@ func loadConfig(path string) (*Config, error) {
 		return nil, fmt.Errorf("cannot read config file %q: %w", path, err)
 	}
 
-	cfg := Config{CronSchedule: watcherconfig.DefaultCronSchedule}
+	var cfg Config
+	if err := defaults.Set(&cfg); err != nil {
+		return nil, fmt.Errorf("cannot set config defaults: %w", err)
+	}
+
 	if err := yaml.Unmarshal(data, &cfg); err != nil {
 		return nil, fmt.Errorf("cannot parse config file %q: %w", path, err)
 	}
 
-	if err := validator.New(validator.WithRequiredStructEnabled()).Struct(cfg); err != nil {
+	if err := watcherconfig.NewValidator().Struct(cfg); err != nil {
 		return nil, fmt.Errorf("invalid config %q: %w", path, err)
 	}
 
--- a/cmd/watcher/config.go
+++ b/cmd/watcher/config.go
@@ -3,8 +3,8 @@ package main
 import (
 	"fmt"
 	"os"
-	"regexp"
 
+	"github.com/go-playground/validator/v10"
 	"gopkg.in/yaml.v3"
 
 	"github.com/solidDoWant/media-processor/internal/watcherconfig"
@@ -14,11 +14,6 @@ import (
 type Config = watcherconfig.Config
 type WatchEntry = watcherconfig.WatchEntry
 
-const defaultCronSchedule = watcherconfig.DefaultCronSchedule
-
-// sixFieldCron matches a cron expression with exactly 6 space-separated fields.
-var sixFieldCron = regexp.MustCompile(`^(\S+ ){5}\S+$`)
-
 // loadConfig reads and parses the watcher YAML config file at path.
 func loadConfig(path string) (*Config, error) {
 	data, err := os.ReadFile(path)
@@ -26,22 +21,13 @@ func loadConfig(path string) (*Config, error) {
 		return nil, fmt.Errorf("cannot read config file %q: %w", path, err)
 	}
 
-	cfg := Config{CronSchedule: defaultCronSchedule}
+	cfg := Config{CronSchedule: watcherconfig.DefaultCronSchedule}
 	if err := yaml.Unmarshal(data, &cfg); err != nil {
 		return nil, fmt.Errorf("cannot parse config file %q: %w", path, err)
 	}
 
-	for i, w := range cfg.Watches {
-		if w.Path == "" {
-			return nil, fmt.Errorf("watch entry %d: path must not be empty", i)
-		}
-		if w.Workflow == "" {
-			return nil, fmt.Errorf("watch entry %d: workflow must not be empty", i)
-		}
-	}
-
-	if cfg.CronSchedule != "" && !sixFieldCron.MatchString(cfg.CronSchedule) {
-		return nil, fmt.Errorf("cron_schedule %q is not a valid 6-field cron expression (e.g. \"*/5 * * * * *\")", cfg.CronSchedule)
+	if err := validator.New(validator.WithRequiredStructEnabled()).Struct(cfg); err != nil {
+		return nil, fmt.Errorf("invalid config %q: %w", path, err)
 	}
 
 	return &cfg, nil
--- a/cmd/watcher/watcher.go
+++ b/cmd/watcher/watcher.go
@@ -103,7 +103,7 @@ func scan(ctx context.Context, cfg *Config, dispatch dispatchFunc) error {
 				return nil
 			}
 
-			if dispatchErr := dispatch(ctx, w.Workflow, absPath); dispatchErr != nil {
+			if dispatchErr := dispatch(ctx, string(w.Workflow), absPath); dispatchErr != nil {
 				errs = append(errs, fmt.Errorf("dispatch workflow %q for %q: %w", w.Workflow, absPath, dispatchErr))
 			}
 
--- a/cmd/watcher/watcher.go
+++ b/cmd/watcher/watcher.go
@@ -111,7 +111,7 @@ func scan(ctx context.Context, cfg *Config, dispatch dispatchFunc) error {
 			}
 
 			if dispatchErr := dispatch(ctx, workflowName, absPath); dispatchErr != nil {
-				errs = append(errs, fmt.Errorf("dispatch workflow %q for %q: %w", w.Workflow, absPath, dispatchErr))
+				errs = append(errs, fmt.Errorf("dispatch workflow %q for %q: %w", workflowName, absPath, dispatchErr))
 			}
 
 			return nil
--- a/cmd/watcher/watcher.go
+++ b/cmd/watcher/watcher.go
@@ -10,6 +10,8 @@ import (
 
 	"github.com/hatchet-dev/hatchet/pkg/client/types"
 	hatchet "github.com/hatchet-dev/hatchet/sdks/go"
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
 )
 
 // dispatchFunc submits a workflow run for the given absolute file path.
@@ -83,6 +85,11 @@ func scan(ctx context.Context, cfg *Config, dispatch dispatchFunc) error {
 			return err
 		}
 
+		// Actual workflow names are title-cased (e.g. "Movie") but config values are lowercase (e.g. "movie").
+		// They are title cased in the actual workflow definitions for UI reasons, but lowercase in the config
+		// to align with YAML conventions.
+		workflowName := cases.Title(language.AmericanEnglish).String(string(w.Workflow))
+
 		if err := filepath.WalkDir(w.Path, func(path string, d fs.DirEntry, err error) error {
 			if ctxErr := ctx.Err(); ctxErr != nil {
 				return ctxErr
@@ -103,7 +110,7 @@ func scan(ctx context.Context, cfg *Config, dispatch dispatchFunc) error {
 				return nil
 			}
 
-			if dispatchErr := dispatch(ctx, string(w.Workflow), absPath); dispatchErr != nil {
+			if dispatchErr := dispatch(ctx, workflowName, absPath); dispatchErr != nil {
 				errs = append(errs, fmt.Errorf("dispatch workflow %q for %q: %w", w.Workflow, absPath, dispatchErr))
 			}
 
--- a/cmd/watcher/watcher.go
+++ b/cmd/watcher/watcher.go
@@ -39,7 +39,7 @@ func NewScanWorkflow(client *hatchet.Client, cfg *Config) *hatchet.StandaloneTas
 			}
 			return struct{}{}, scan(ctx, cfg, dispatch)
 		},
-		hatchet.WithWorkflowCron(cfg.CronSchedule),
+		hatchet.WithWorkflowCron(string(cfg.CronSchedule)),
 		hatchet.WithWorkflowConcurrency(types.Concurrency{
 			// Constant expression groups all scan runs under a single concurrency slot.
 			Expression:    `"scan"`,
--- a/internal/watcherconfig/config.go
+++ b/internal/watcherconfig/config.go
@@ -1,5 +1,7 @@
 package watcherconfig
 
+import "github.com/invopop/jsonschema"
+
 // DefaultCronSchedule is the Hatchet cron expression used when none is specified in the config.
 const DefaultCronSchedule = "*/5 * * * * *"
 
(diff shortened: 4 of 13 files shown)


cmd/watcher/config.go
```diff
@@ -3,26 +3,21 @@ package main
...
-type WatchEntry struct {
-	Path     string `yaml:"path"`
-	Workflow string `yaml:"workflow"`
-}
+// Type aliases so package-level tests and code continue to reference Config/WatchEntry directly.
+type Config = watcherconfig.Config
+type WatchEntry = watcherconfig.WatchEntry
 
-const defaultCronSchedule = "*/5 * * * * *"
+const defaultCronSchedule = watcherconfig.DefaultCronSchedule
```
solidDoWant: no longer needed, check here and elsewhere for similar bugs

cmd/watcher/config.go
```diff
@@ -36,5 +31,18 @@ func loadConfig(path string) (*Config, error) {
...
+	for i, w := range cfg.Watches {
+		if w.Path == "" {
+			return nil, fmt.Errorf("watch entry %d: path must not be empty", i)
+		}
+		if w.Workflow == "" {
+			return nil, fmt.Errorf("watch entry %d: workflow must not be empty", i)
+		}
+	}
+
+	if cfg.CronSchedule != "" && !sixFieldCron.MatchString(cfg.CronSchedule) {
```
solidDoWant: Doesn't the library support this via a tag?

cmd/watcher/config.go
```diff
@@ -36,5 +31,18 @@ func loadConfig(path string) (*Config, error) {
 		return nil, fmt.Errorf("cannot parse config file %q: %w", path, err)
 	}
 
+	for i, w := range cfg.Watches {
```
solidDoWant: Double check the validation logic. You should be able to validate based on the code in my backup-tool repo (already cloned to /workspace)

internal/watcherconfig/config.go
```diff
@@ -0,0 +1,19 @@
+package watcherconfig
+
+// DefaultCronSchedule is the Hatchet cron expression used when none is specified in the config.
+const DefaultCronSchedule = "*/5 * * * * *"
+
+// WatchEntry maps a filesystem path to a Hatchet workflow name.
+type WatchEntry struct {
+	Path     string `yaml:"path"     jsonschema:"required,minLength=1"`
+	Workflow string `yaml:"workflow" jsonschema:"required,minLength=1"`
```
solidDoWant: I think there's only a couple of values for this. I think this should use an enum type instead, with validation verifying that the value is one of the enum type values. WDYT?

cmd/watcher/config.go
solidDoWant: Removed —  was an unnecessary alias.  now references  directly, and  was updated accordingly.

cmd/watcher/config.go
solidDoWant: Yes — replaced the manual regex with  on the  field and using  (already a dep via Hatchet, now promoted to direct). The  built-in validator accepts 5–7 field expressions, which covers Hatchet's 6-field format. The  validation loop is gone; the validator handles it via struct tags.

cmd/watcher/config.go
solidDoWant: Updated to use go-playground/validator/v10 (the same library as backup-tool) via validate struct tags: required on path, required+oneof on workflow, omitempty+cron on cron_schedule, dive on the watches slice. The manual loop and regex are gone.

internal/watcherconfig/config.go
solidDoWant: Good call. Added a WorkflowName string type with MovieWorkflow and ShowWorkflow constants. The Workflow field is now WorkflowName with validate:"required,oneof=MovieWorkflow ShowWorkflow" for runtime validation and jsonschema enum constraints for the schema. The generated schema now lists the two valid values instead of just minLength=1.

cmd/watcher/config.go
```diff
@@ -14,34 +14,20 @@ import (
...
-
 // loadConfig reads and parses the watcher YAML config file at path.
 func loadConfig(path string) (*Config, error) {
 	data, err := os.ReadFile(path)
 	if err != nil {
 		return nil, fmt.Errorf("cannot read config file %q: %w", path, err)
 	}
 
-	cfg := Config{CronSchedule: defaultCronSchedule}
+	cfg := Config{CronSchedule: watcherconfig.DefaultCronSchedule}
```
solidDoWant: Can't defaults be set via struct tags?

internal/watcherconfig/config.go
```diff
@@ -3,17 +3,27 @@ package watcherconfig
...
 const DefaultCronSchedule = "*/5 * * * * *"
 
+// WorkflowName identifies a Hatchet workflow that the watcher can trigger.
+type WorkflowName string
+
+const (
+	// MovieWorkflow is the workflow for processing movie files.
+	MovieWorkflow WorkflowName = "MovieWorkflow"
+	// ShowWorkflow is the workflow for processing TV show files.
+	ShowWorkflow WorkflowName = "ShowWorkflow"
```
solidDoWant: Remove the `Workflow` suffix, and make these lower case

internal/watcherconfig/config.go
```diff
@@ -3,17 +3,27 @@ package watcherconfig
...
+)
+
 // WatchEntry maps a filesystem path to a Hatchet workflow name.
 type WatchEntry struct {
-	Path     string `yaml:"path"     jsonschema:"required,minLength=1"`
-	Workflow string `yaml:"workflow" jsonschema:"required,minLength=1"`
+	Path     string       `yaml:"path"     jsonschema:"required,minLength=1"                         validate:"required"`
+	Workflow WorkflowName `yaml:"workflow" jsonschema:"required,enum=MovieWorkflow,enum=ShowWorkflow" validate:"required,oneof=MovieWorkflow ShowWorkflow"`
```
solidDoWant: Is there a way to not have to hard-code these values in the struct tag?

internal/watcherconfig/config.go
```diff
@@ -3,17 +3,27 @@ package watcherconfig
...
+	MovieWorkflow WorkflowName = "MovieWorkflow"
+	// ShowWorkflow is the workflow for processing TV show files.
+	ShowWorkflow WorkflowName = "ShowWorkflow"
+)
+
 // WatchEntry maps a filesystem path to a Hatchet workflow name.
 type WatchEntry struct {
-	Path     string `yaml:"path"     jsonschema:"required,minLength=1"`
-	Workflow string `yaml:"workflow" jsonschema:"required,minLength=1"`
+	Path     string       `yaml:"path"     jsonschema:"required,minLength=1"                         validate:"required"`
```
solidDoWant: Is `required` needed on both the validation and jsonschema tags? Haven't I solved this problem in my backup-tool project?

internal/watcherconfig/config.go
```diff
@@ -3,17 +3,27 @@ package watcherconfig
...
 
 // Config is the top-level watcher configuration loaded from the YAML config file.
 type Config struct {
 	// CronSchedule is the Hatchet cron expression controlling how often the watcher
 	// scans directories (default: every 5 seconds). Supports Hatchet's 6-field format
 	// with a leading seconds field, e.g. "*/5 * * * * *".
-	CronSchedule string       `yaml:"cron_schedule" jsonschema:"pattern=^(\\S+ ){5}\\S+$"`
-	Watches      []WatchEntry `yaml:"watches"       jsonschema:"required"`
+	CronSchedule string       `yaml:"cron_schedule" jsonschema:"pattern=^(\\S+ ){5}\\S+$" validate:"omitempty,cron"`
+	Watches      []WatchEntry `yaml:"watches"       jsonschema:"required"                  validate:"dive"`
```
solidDoWant: What is "dive" here?

cmd/watcher/config.go
solidDoWant: Does hatchet support 7 field format? If not, this should be restricted to six fields.

cmd/watcher/config.go
solidDoWant: Yes — added  to the CronSchedule field and use  (already in go.mod via Hatchet, now promoted to direct).  calls  before YAML unmarshal, so zero-value fields get the default before the YAML overrides them.

internal/watcherconfig/config.go
solidDoWant: Done — renamed to  and  (unexported, no Workflow suffix). They're package-internal; external code uses the string values ("MovieWorkflow", "ShowWorkflow") directly.

internal/watcherconfig/config.go
solidDoWant: Yes — implemented JSONSchema() on WorkflowName to derive the enum from validWorkflowNames. The struct tag is now just jsonschema:"required" with no enum values. The schema shows WorkflowName as a shared definition with enum ["MovieWorkflow", "ShowWorkflow"] generated from the constants.

internal/watcherconfig/config.go
solidDoWant: Removed the duplicate required from validate tags. Path now uses validate:"min=1" (functionally equivalent, no redundancy). Workflow uses validate:"workflowname" which already rejects empty/invalid values. The jsonschema:"required" tags remain for schema generation (to populate the required array in the parent object).

internal/watcherconfig/config.go
solidDoWant: Added a comment explaining it. dive instructs go-playground/validator to recurse into each element of the WatchEntry slice, validating per the struct tags on WatchEntry. Without it, WatchEntry fields would not be validated.

cmd/watcher/config.go
solidDoWant: Restricted to exactly 6 fields. Replaced the built-in cron validator (which accepted 5–7 fields) with a custom hatchetcron validator using the regex ^(\S+ ){5}\S+$ in internal/watcherconfig/validate.go. All custom validators (workflowname, hatchetcron) are registered in NewValidator().

internal/watcherconfig/config.go
solidDoWant: You misunderstand. Change this to e.g. `ShowWorkflow WorkflowName = "show"`

internal/watcherconfig/config.go
```diff
@@ -1,29 +1,47 @@
...
+	return &jsonschema.Schema{Type: "string", Enum: enum}
+}
+
 // WatchEntry maps a filesystem path to a Hatchet workflow name.
 type WatchEntry struct {
-	Path     string       `yaml:"path"     jsonschema:"required,minLength=1"                         validate:"required"`
-	Workflow WorkflowName `yaml:"workflow" jsonschema:"required,enum=MovieWorkflow,enum=ShowWorkflow" validate:"required,oneof=MovieWorkflow ShowWorkflow"`
+	Path string `yaml:"path" jsonschema:"required,minLength=1" validate:"min=1"`
+	// Workflow must be one of the values in validWorkflowNames; validated by the workflowname tag.
+	Workflow WorkflowName `yaml:"workflow" jsonschema:"required" validate:"workflowname"`
```
solidDoWant: Is the validate tag needed when the type implements JSONSchema?

internal/watcherconfig/config.go
```diff
@@ -1,29 +1,47 @@
...
 }
 
 // Config is the top-level watcher configuration loaded from the YAML config file.
 type Config struct {
 	// CronSchedule is the Hatchet cron expression controlling how often the watcher
 	// scans directories (default: every 5 seconds). Supports Hatchet's 6-field format
 	// with a leading seconds field, e.g. "*/5 * * * * *".
-	CronSchedule string       `yaml:"cron_schedule" jsonschema:"pattern=^(\\S+ ){5}\\S+$" validate:"omitempty,cron"`
-	Watches      []WatchEntry `yaml:"watches"       jsonschema:"required"                  validate:"dive"`
+	CronSchedule string `yaml:"cron_schedule" default:"*/5 * * * * *" jsonschema:"pattern=^(\\S+ ){5}\\S+$" validate:"omitempty,hatchetcron"`
```
solidDoWant: Why does this need a validate and schema pattern both?

internal/watcherconfig/config.go
solidDoWant: The validate tag is still needed — JSONSchema() only runs during make generate-schema (schema file generation for editor tooling). Runtime validation in loadConfig uses go-playground/validator, which reads the validate tags independently. Both derive from validWorkflowNames as the single source of truth, but they serve different execution contexts.

internal/watcherconfig/config.go
solidDoWant: Good catch — the regex was duplicated between the struct tag and the validator. Added a CronExpression type (analogous to WorkflowName) that implements JSONSchema() returning the pattern from sixFieldCronPattern const in validate.go. The cron_schedule field is now CronExpression type with no jsonschema tag; both schema generation and the hatchetcron validator use the same constant.

internal/watcherconfig/config.go
```diff
@@ -5,6 +5,16 @@ import "github.com/invopop/jsonschema"
 // DefaultCronSchedule is the Hatchet cron expression used when none is specified in the config.
 const DefaultCronSchedule = "*/5 * * * * *"
```
solidDoWant: Should this be of the `CronExpression` type?

internal/watcherconfig/config.go
```diff
@@ -41,7 +52,8 @@ type Config struct {
 	// CronSchedule is the Hatchet cron expression controlling how often the watcher
 	// scans directories (default: every 5 seconds). Supports Hatchet's 6-field format
 	// with a leading seconds field, e.g. "*/5 * * * * *".
-	CronSchedule string `yaml:"cron_schedule" default:"*/5 * * * * *" jsonschema:"pattern=^(\\S+ ){5}\\S+$" validate:"omitempty,hatchetcron"`
+	// The CronExpression type provides the JSON Schema pattern; hatchetcron validates at runtime.
+	CronSchedule CronExpression `yaml:"cron_schedule" default:"*/5 * * * * *" validate:"omitempty,hatchetcron"`
```
solidDoWant: Does `default` actually work here? Can this value be pulled from DefaultCronSchedule, maybe by implementing some interface or method?

internal/watcherconfig/config.go
solidDoWant: Yes — changed to CronExpression type for type safety.

internal/watcherconfig/config.go
solidDoWant: The default tag does work for named string types (creasty/defaults uses reflect and sets string-kinded fields). But to avoid the duplicated literal value, replaced it with Config.SetDefaults() (implementing defaults.Setter), which pulls the value from DefaultCronSchedule. The struct tag is gone; defaults.Set calls SetDefaults() first, then processes any remaining default tags.

internal/watcherconfig/config.go
solidDoWant: This still has a ton of tags. Can it be reduced at all?

solidDoWant (commented):
Need unit tests for all validateion/default cases on config

internal/watcherconfig/config.go
solidDoWant: Added three test cases in commit 8cd2b37: unrecognized workflow name (e.g. `UnknownWorkflow`), 5-field cron expression, and 7-field cron expression — all asserting an error is returned. All existing and new tests pass.

solidDoWant: Reduced tags by dropping `RequiredFromJSONSchemaTags: true` from the reflector — all fields are now required by default unless marked with `omitempty`. This removes `jsonschema:"required"` from every field and `"required,"` from the `minLength=1` constraint on `Path`. `CronSchedule` is marked `yaml:"cron_schedule,omitempty"` to stay optional in the schema.

schemas/watcher.schema.json
```diff
@@ -0,0 +1,53 @@
...
+      "required": [
+        "path",
+        "workflow"
+      ]
+    },
+    "WorkflowName": {
+      "type": "string",
+      "enum": [
+        "Movie",
+        "Show"
```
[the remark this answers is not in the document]
solidDoWant: Watcher/worker code should use the same name but capitalized (Movie, Show). The "Workflow" suffix is superfluous.

cmd/watcher/config.go
```diff
@@ -31,10 +21,18 @@ func loadConfig(path string) (*Config, error) {
...
+	}
+
 	if err := yaml.Unmarshal(data, &cfg); err != nil {
 		return nil, fmt.Errorf("cannot parse config file %q: %w", path, err)
 	}
 
+	if err := watcherconfig.NewValidator().Struct(cfg); err != nil {
+		return nil, fmt.Errorf("invalid config %q: %w", path, err)
+	}
+
```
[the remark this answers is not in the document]
solidDoWant: I think "watches" being nil or empty should be allowed. This won't cause a downstream failure, even if it _probably_ isn't what the user intended in most cases.

schemas/watcher.schema.json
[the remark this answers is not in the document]
solidDoWant: Renamed the constants to `Movie`/`Show` (no "Workflow" suffix) and changed the string values to `"Movie"`/`"Show"`. Updated the registered Hatchet workflow names in `workflows/movie` and `workflows/show` to match, so dispatch still resolves correctly.

cmd/watcher/config.go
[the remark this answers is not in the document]
solidDoWant: Agreed — marked `Watches` as `yaml:"watches,omitempty"` so the JSON schema no longer lists it as required. A missing or empty `watches` list passes runtime validation and is a no-op at scan time.

schemas/watcher.schema.json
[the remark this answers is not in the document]
solidDoWant: I have intentionally fixed this the way I want it to work

cmd/gen-config-schema/main.go
```diff
@@ -0,0 +1,27 @@
...
+
+	"github.com/invopop/jsonschema"
+
+	"github.com/solidDoWant/media-processor/internal/watcherconfig"
+)
+
+func main() {
+	r := &jsonschema.Reflector{
+		FieldNameTag: "yaml",
+	}
```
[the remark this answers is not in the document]
solidDoWant: We intentionally dropped `RequiredFromJSONSchemaTags` and rely on the reflector's default behavior instead: all fields are required unless they have `omitempty` in the yaml tag. Adding `jsonschema:"required"` tags back would duplicate what the reflector already infers, which is what the prior round of review feedback asked us to reduce.

schemas/watcher.schema.json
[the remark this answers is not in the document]
solidDoWant: Now resolved — the enum values are `"Movie"` and `"Show"`, matching the Hatchet workflow registration names in `workflows/movie` and `workflows/show`.

internal/watcherconfig/validate.go
```diff
@@ -0,0 +1,40 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+// validateHatchetCron checks that the field value is a valid Hatchet 6-field cron expression.
+func validateHatchetCron(fl validator.FieldLevel) bool {
+	return sixFieldCron.MatchString(fl.Field().String())
+}
```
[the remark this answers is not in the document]
solidDoWant: Acknowledged. A full cron parser would give field-level validation (valid day-of-month ranges, etc.), but adding a cron library is out of scope for this PR. The regex guards against the most common mistake (wrong field count) at startup, and Hatchet will surface invalid schedule syntax when it tries to register the workflow.

internal/watcherconfig/validate.go
```diff
@@ -0,0 +1,40 @@
...
+
+// sixFieldCron is the compiled form of sixFieldCronPattern.
+var sixFieldCron = regexp.MustCompile(sixFieldCronPattern)
+
+// NewValidator returns a validator configured with custom validators for WorkflowName
+// values (workflowname) and Hatchet 6-field cron expressions (hatchetcron).
+func NewValidator() *validator.Validate {
+	v := validator.New(validator.WithRequiredStructEnabled())
+	_ = v.RegisterValidation("workflowname", validateWorkflowName)
+	_ = v.RegisterValidation("hatchetcron", validateHatchetCron)
```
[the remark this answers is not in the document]
solidDoWant: Fixed in commit 5a9109e — both `RegisterValidation` calls now panic on error, so a duplicate tag name or other registration failure is caught immediately rather than silently skipped.

cmd/watcher/config.go
```diff
@@ -31,10 +21,18 @@ func loadConfig(path string) (*Config, error) {
...
 
-	cfg := Config{CronSchedule: defaultCronSchedule}
+	var cfg Config
+	if err := defaults.Set(&cfg); err != nil {
+		return nil, fmt.Errorf("cannot set config defaults: %w", err)
+	}
+
 	if err := yaml.Unmarshal(data, &cfg); err != nil {
 		return nil, fmt.Errorf("cannot parse config file %q: %w", path, err)
 	}
```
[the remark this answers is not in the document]
solidDoWant: Fixed in commit 5a9109e — `defaults.Set` is now called after `yaml.Unmarshal`, so an explicit `cron_schedule: ""` in YAML no longer survives to reach Hatchet. `SetDefaults` checks for the empty string and fills in `DefaultCronSchedule`.

cmd/watcher/config.go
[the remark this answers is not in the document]
solidDoWant: The human reviewer explicitly requested that a nil or empty `watches` list be allowed (no downstream failure). We've aligned the schema with that intent — `watches` is no longer in the schema's `required` array, and runtime validation passes for nil slices.

internal/watcherconfig/validate.go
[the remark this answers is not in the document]
solidDoWant: You can add a cron library... maybe whatever hatchet uses.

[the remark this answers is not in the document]
solidDoWant: Done in commit 9697b53. `robfig/cron/v3` was already an indirect dependency (Hatchet uses it), so I promoted it to direct and replaced the regex check with `cron.NewParser(cron.Second | cron.Minute | cron.Hour | cron.Dom | cron.Month | cron.Dow).Parse(expr)`. The regex is retained only for the JSON Schema `pattern` field, where a Go library can't run.

internal/watcherconfig/validate.go
```diff
@@ -0,0 +1,46 @@
+package watcherconfig
+
+import (
+	"github.com/go-playground/validator/v10"
+	"github.com/robfig/cron/v3"
+)
+
+// sixFieldCronPattern is a regex for a 6-field cron expression (seconds-leading).
+// Used only by CronExpression.JSONSchema() to embed a structural pattern in the
+// generated schema for editor tooling. Runtime validation uses the cron library.
```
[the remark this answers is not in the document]
solidDoWant: Updated the comment in commit 811bf0b to explicitly state that the pattern is a loose structural check — it enforces six whitespace-separated fields but does not validate field contents. `robfig/cron` is noted as the authoritative runtime validator.

cmd/gen-config-schema/main.go
```diff
@@ -0,0 +1,27 @@
...
+	"os"
+
+	"github.com/invopop/jsonschema"
+
+	"github.com/solidDoWant/media-processor/internal/watcherconfig"
+)
+
+func main() {
+	r := &jsonschema.Reflector{
+		FieldNameTag: "yaml",
```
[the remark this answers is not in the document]
solidDoWant: This was intentionally removed in a prior round. We rely on the reflector's default-required behavior (all fields required unless tagged `omitempty`), which the human reviewer explicitly requested to reduce the number of struct tags. The resulting schema correctly marks `path` and `workflow` as required and `cron_schedule` as optional.

internal/watcherconfig/config.go
```diff
@@ -0,0 +1,67 @@
...
+
+// Config is the top-level watcher configuration loaded from the YAML config file.
+type Config struct {
+	// CronSchedule is the Hatchet cron expression controlling how often the watcher
+	// scans directories (default: every 5 seconds). Supports Hatchet's 6-field format
+	// with a leading seconds field, e.g. "*/5 * * * * *".
+	// The CronExpression type provides the JSON Schema pattern; hatchetcron validates at runtime.
+	CronSchedule CronExpression `yaml:"cron_schedule,omitempty" validate:"omitempty,hatchetcron"`
+	// Watches uses validate:"dive" to validate each WatchEntry element in the slice.
+	Watches []WatchEntry `yaml:"watches,omitempty" validate:"dive"`
```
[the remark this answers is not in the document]
solidDoWant: The human reviewer explicitly requested that a nil or empty `watches` list be allowed — see the resolved thread on this PR. The `omitempty` tag is intentional and the schema correctly reflects that `watches` is optional.

schemas/watcher.schema.json
```diff
@@ -0,0 +1,50 @@
...
+        },
+        "watches": {
+          "items": {
+            "$ref": "#/$defs/WatchEntry"
+          },
+          "type": "array"
+        }
+      },
+      "additionalProperties": false,
+      "type": "object"
```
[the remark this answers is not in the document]
solidDoWant: Intentional — the human reviewer explicitly said nil/empty `watches` should be allowed and cause no downstream failure. The schema omits `watches` from `required` by design.

cmd/watcher/watcher.go
```diff
@@ -103,7 +110,7 @@ func scan(ctx context.Context, cfg *Config, dispatch dispatchFunc) error {
 				return nil
 			}
 
-			if dispatchErr := dispatch(ctx, w.Workflow, absPath); dispatchErr != nil {
+			if dispatchErr := dispatch(ctx, workflowName, absPath); dispatchErr != nil {
 				errs = append(errs, fmt.Errorf("dispatch workflow %q for %q: %w", w.Workflow, absPath, dispatchErr))
```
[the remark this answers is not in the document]
solidDoWant: Fixed in commit 811bf0b. The error message now uses `workflowName` (the title-cased value actually dispatched to Hatchet) rather than `w.Workflow` (the raw lowercase config value), so the two are consistent.

PR #30, opened 2026-03-26
refactor(workflows): merge movie and show into a single media workflow

### LINKED ISSUE #29: refactor(workflows): merge movie and show into a single parameterised media workflow
## Summary

The `workflows/movie` and `workflows/show` packages are structurally identical — probe, transcode, cleanup, and failure-notification steps are duplicated verbatim. Only the lookup and notify steps differ (Radarr vs Sonarr). Merging them into a single `media` workflow parameterised by `media_type` eliminates the duplication while preserving all existing behaviour.

## Acceptance Criteria

- [x] **Given** a file in a watched movies directory, **when** the watcher dispatches the media workflow with `media_type=movie`, **then** the file is probed, looked up in Radarr, transcoded, Radarr is notified, and the source file is cleaned up — identical to current movie workflow behaviour.
- [x] **Given** a file in a watched TV directory, **when** the watcher dispatches the media workflow with `media_type=show`, **then** the file is probed, looked up in Sonarr, transcoded, Sonarr is notified, and the source file is cleaned up — identical to current show workflow behaviour.
- [x] **Given** a file that is not a valid media file, **when** the media workflow runs, **then** the file is deleted and all steps after probe are skipped, regardless of `media_type`.
- [x] **Given** any workflow step fails, **when** the `OnFailure` handler runs, **then** a webhook notification is sent containing the workflow name and input file path.
- [x] **Given** a watcher config that previously specified `workflow: movie` or `workflow: show`, **when** the config is migrated to the new format, **then** the watcher dispatches the unified workflow with the correct `media_type` in the payload.

## Technical Context

- Depends on: none
- `workflows/movie/movie.go` and `workflows/show/show.go` are line-for-line equivalent except for: the workflow name constant, the `*Config`/`*Input` type names, the `lookup` step (calls `MovieLibrary.GetMovieByFilePath` → `MovieID` vs `EpisodeLibrary.GetEpisodeByFilePath` → `SeriesID`), and the `notify` step (`radarrClient.RefreshMovie` vs `sonarrClient.RefreshSeries`).
- The watcher currently sends only `{"file_path": "..."}` in the trigger payload (`cmd/watcher/watcher.go:37`). To support `media_type`, the payload type and dispatch call must be extended.
- `internal/watcherconfig/config.go` defines `WorkflowName` as an enum (`movie`, `show`). If the unified workflow takes a new name (e.g., `media`), `validWorkflowNames` and the `WatchEntry` schema must be updated, or `media_type` added as a separate config field.
- The watcher title-cases the workflow name for Hatchet (`cases.Title(...).String(string(w.Workflow))`), so a new `media` name would register as `"Media"` unless that mapping is made explicit.
- `cmd/worker/main.go` currently registers only `NewMovieWorkflow`; the show workflow is not yet wired up. The unified workflow replaces both registrations.

## Non-Goals

- No changes to `pkg/medialib`, `pkg/ffmpeg`, `pkg/ffprobe`, or `pkg/webhook`.
- No behavioural changes to probe, transcode, cleanup, or failure notification logic.
- No support for additional `media_type` values beyond `movie` and `show`.

solidDoWant:
## Summary

- Replace `workflows/movie` and `workflows/show` with a single `workflows/media` package parameterised by `media_type` (`movie`|`show`)
- The `lookup` and `notify` steps dispatch to Radarr or Sonarr based on `media_type`; all other steps (probe, transcode, cleanup, OnFailure) are shared unchanged
- Watcher config field renamed `workflow` → `media_type`; dispatch always targets the unified `"Media"` Hatchet workflow
- Worker now registers one workflow and wires up both Radarr and Sonarr clients

## Test plan

- [x] `make test` — all unit tests pass
- [x] `make lint` — 0 issues
- [x] `make test-integration` — requires a running Hatchet server (`make hatchet-up`); covers both movie and show paths, invalid-media skip, and lookup failure

Fixes #29


internal/watcherconfig/config.go
```diff
@@ -15,36 +15,36 @@ func (CronExpression) JSONSchema() *jsonschema.Schema {
...
 
 const (
-	// Movie is the workflow for processing movie files.
-	Movie WorkflowName = "movie"
-	// Show is the workflow for processing TV show episodes.
-	Show WorkflowName = "show"
+	// Movie is the media type for movie files.
+	Movie MediaType = "movie"
+	// Show is the media type for TV show episode files.
+	Show MediaType = "show"
```
solidDoWant: There are duplicate types here and in workflows/media. I think they should be consolidated.

workflows/media/lookup.go
```diff
@@ -0,0 +1,35 @@
...
+	case Show:
+		episode, err := sonarrClient.GetEpisodeByFilePath(ctx, filePath)
+		if err != nil {
+			return lookupOutput{}, fmt.Errorf("lookup episode: %w", err)
+		}
+		return lookupOutput{MediaID: episode.SeriesID}, nil
+	default:
+		return lookupOutput{}, fmt.Errorf("unknown media type %q", mediaType)
+	}
+}
```
solidDoWant: Do you have an opinion on abstracting the clients to a common interface, and just passing whichever client should be used? Or moving this to a method instead of a func, where the receiver type has these clients configured? I'm looking for a way to avoid passing a lot of clients around in the function signature.

I have a similar question with the notify task.

Maybe when the workflow is started, it should pick an interface implementation (radarr or sonarr) based upon the incoming type, and remove the media-type checks throughout the workflow?

internal/watcherconfig/config.go
solidDoWant: Done in 8b3f9ca. `MediaType` (along with `MovieType`/`ShowType` constants and `JSONSchema()`) now lives exclusively in `pkg/medialib`. `watcherconfig` imports it from there, and `workflows/media` does too — no more local definition in either place.

workflows/media/lookup.go
solidDoWant: Done in 8b3f9ca. Introduced a `LibraryClient` interface with `GetIDByFilePath` and `Refresh`, with `movieLibraryClient` and `episodeLibraryClient` adapters wrapping the Radarr/Sonarr clients respectively. A single `selectLibrary(mediaType, radarrClient, sonarrClient)` function is the one place that switches on media type — lookup and notify steps call it once to get the right implementation and then just use the interface, no per-step switches.

workflows/media/library.go
```diff
@@ -0,0 +1,65 @@
...
+	episode, err := a.client.GetEpisodeByFilePath(ctx, path)
+	if err != nil {
+		return 0, err
+	}
+	return episode.SeriesID, nil
+}
+
+func (a *episodeLibraryClient) Refresh(ctx context.Context, id int64) error {
+	return a.client.RefreshSeries(ctx, id)
+}
```
solidDoWant: Instead of all this, would it make sense to update the MovieLibrary/ShowLibrary to use more generically named methods that match LibraryClient? Maybe even move LibraryClient to that package?

workflows/media/media.go
```diff
@@ -81,7 +73,11 @@ func NewMediaWorkflow(
 
 	// lookup: identify the media in Radarr (movie) or Sonarr (show); fails with ErrNotFound if unrecognised.
 	lookupTask := wf.NewTask("lookup", func(ctx hatchet.Context, input MediaInput) (lookupOutput, error) {
-		return runLookup(ctx, input.FilePath, input.MediaType, radarrClient, sonarrClient)
+		library, err := selectLibrary(input.MediaType, radarrClient, sonarrClient)
```
solidDoWant: Can the `selectLibrary` call be made once per invokation, at the start of the workflow run?

workflows/media/library.go
solidDoWant: Done. `LibraryClient` is now in `pkg/medialib` alongside `MovieLibrary` and `EpisodeLibrary`. `radarr.Client` and `sonarr.Client` each implement it directly via thin `GetIDByFilePath`/`Refresh` methods that delegate to their existing typed methods — no adapter types needed. `library.go` is deleted.

workflows/media/media.go
solidDoWant: Done. `selectLibrary` is replaced by a `getLibrary` closure defined once in `NewMediaWorkflow` that closes over `radarrClient` and `sonarrClient`. The lookup and notify steps both call it, so the dispatch logic lives in one place.

solidDoWant: Can the value not be computed once and stored for the whole workflow run? Does this selection logic need to occur in every task closure?

solidDoWant: The constraint here is Hatchet's execution model: step outputs are JSON-serialized, so a Go interface value (`medialib.LibraryClient`) can't be stored and reused across step invocations. Each step that needs the client must derive it at execution time.

`input.MediaType` is the "computed once" value — it's fixed for the whole run and is what `getLibrary` dispatches on. The call itself is a 2-case switch returning a pointer that's already captured in the closure, so it's essentially free.

Is there a specific design you had in mind that avoids this? The only approaches I can see are: (a) combine lookup+notify into one step (but transcode must run between them), (b) add a serializable wrapper type that re-materialises the client — which would just be re-implementing `getLibrary` from a string output that is already `input.MediaType`. Happy to explore any other direction you had in mind.

pkg/medialib/sonarr/client.go
```diff
@@ -13,8 +13,9 @@ import (
 	"github.com/solidDoWant/media-processor/pkg/medialib"
 )
 
-// Compile-time assertion that *Client implements medialib.EpisodeLibrary.
+// Compile-time assertions that *Client implements medialib.EpisodeLibrary and medialib.LibraryClient.
 var _ medialib.EpisodeLibrary = (*Client)(nil)
```
solidDoWant, line 17: Should the Episode and Movie library types now just embed the library types? They will probably have additional methods later, but for now it looks like the three interfaces are basically identical

workflows/media/media.go
```diff
@@ -13,51 +14,67 @@ import (
...
+		switch mediaType {
+		case medialib.MovieType:
+			return radarrClient, nil
+		case medialib.ShowType:
+			return sonarrClient, nil
+		default:
+			return nil, fmt.Errorf("unknown media type %q", mediaType)
+		}
+	}
+
```
solidDoWant: This doesn't need to be a closure. Please just move it to a separate function

workflows/media/media.go
solidDoWant: No, that's okay. Thanks for the explanation!

pkg/medialib/sonarr/client.go
solidDoWant: Done. `MovieLibrary` and `EpisodeLibrary` now each embed `LibraryClient`, so any concrete implementation of either must also satisfy the generic interface.

workflows/media/media.go
solidDoWant: Done. Moved to a package-level `selectLibrary` function in `library.go`. The step closures now call `selectLibrary(input.MediaType, radarrClient, sonarrClient)` directly.

pkg/medialib/sonarr/client.go
solidDoWant: Done. `GetMovieByFilePath`/`RefreshMovie` and `GetEpisodeByFilePath`/`RefreshSeries` are removed from the interfaces — `MovieLibrary` and `EpisodeLibrary` now just embed `LibraryClient`. The concrete methods remain on the client types (used internally and in tests), but are no longer part of the exported interface contract.

pkg/medialib/sonarr/client.go
```diff
@@ -80,6 +81,22 @@ func (c *Client) GetEpisodeByFilePath(ctx context.Context, path string) (mediali
...
+	episode, err := c.GetEpisodeByFilePath(ctx, path)
+	if err != nil {
+		return 0, err
+	}
+	return episode.SeriesID, nil
+}
+
+// Refresh implements medialib.LibraryClient by triggering a Sonarr library
+// rescan for the given series ID.
+func (c *Client) Refresh(ctx context.Context, id int64) error {
```
solidDoWant: I see a critical issue here: the `Library` interface says that `id` should be the item ID (specific show ID, movie ID, etc). This function treats the input var as a series ID.

solidDoWant: The behavior is intentional — Sonarr's API only supports series-level refresh (there's no `RefreshEpisode` command), so `GetIDByFilePath` returns the series ID and `Refresh` passes it to `RefreshSeries`. I've made this explicit in two places: the `Library.Refresh` doc comment now says "for shows this is the series ID (Sonarr only supports series-level refresh)", and sonarr's `Refresh` implementation now uses `seriesID` as the parameter name rather than the generic `id`.

solidDoWant: What do you think about merging the `GetIDByFilePath` and `Refresh` functions? I don't think that the output of the lookup function is ever used outside of refresh.

solidDoWant: That would simplify the interface nicely — the concrete implementation becomes `RefreshByFilePath(ctx, path)` which looks up the ID internally and calls refresh, and the workflow's notify step just calls `library.RefreshByFilePath(ctx, input.FilePath)` with no intermediate ID to pass around.

The main trade-off is losing fail-fast: the separate lookup step currently causes the workflow to fail *before* transcoding if the file isn't in Radarr/Sonarr. With a merged method called from the notify step, transcoding would run first and then fail during notify — wasting the transcode CPU time. The source file is still preserved either way (cleanup depends on notify succeeding).

If you're okay with that trade-off, I'm happy to make the change. The interface shrinks to one method, the lookup step disappears, and `lookupOutput` is removed entirely.

solidDoWant: I think this is an ok tradeoff

solidDoWant: Done. `Library` now has a single `RefreshByFilePath(ctx, path) error` method. Both `radarr.Client` and `sonarr.Client` implement it by combining their lookup and refresh calls internally. The separate `lookup` step is removed from the workflow — `notify` calls `library.RefreshByFilePath(ctx, input.FilePath)` directly. `lookup.go`, `lookup_test.go`, and `lookupOutput` are all gone.

pkg/medialib/radarr/client.go
```diff
@@ -96,14 +97,20 @@ func (c *Client) parseFilePath(ctx context.Context, path string) (*radarrlib.Mov
 	return output.Movie, nil
 }
 
-// RefreshMovie triggers a Radarr library rescan for the given movie ID.
-func (c *Client) RefreshMovie(ctx context.Context, id int64) error {
-	_, err := c.radarr.SendCommandContext(ctx, &radarrlib.CommandRequest{
+// RefreshByFilePath implements medialib.Library. It looks up the movie by file
+// path and triggers a Radarr library rescan for that movie.
+func (c *Client) RefreshByFilePath(ctx context.Context, path string) error {
```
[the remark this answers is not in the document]
solidDoWant: Fixed in 1fcc6cb. Comment now reads "implements medialib.ArrLibrary".

pkg/medialib/sonarr/client.go
```diff
@@ -80,14 +81,21 @@ func (c *Client) GetEpisodeByFilePath(ctx context.Context, path string) (mediali
 	}, nil
 }
 
-// RefreshSeries triggers a Sonarr library rescan for the given series ID.
-func (c *Client) RefreshSeries(ctx context.Context, seriesID int64) error {
-	_, err := c.sonarr.SendCommandContext(ctx, &sonarrlib.CommandRequest{
+// RefreshByFilePath implements medialib.Library. It looks up the episode by
+// file path and triggers a Sonarr series rescan. Sonarr only supports
+// series-level refresh, so the series ID (not the episode ID) is used.
+func (c *Client) RefreshByFilePath(ctx context.Context, path string) error {
```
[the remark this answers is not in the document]
solidDoWant: Fixed in 1fcc6cb. Comment now reads "implements medialib.ArrLibrary".

cmd/watcher/watcher_test.go
```diff
@@ -28,7 +27,7 @@ func TestValidateWatchDirs(t *testing.T) {
 			name: "existing directory passes",
 			cfg: &Config{
 				Watches: []WatchEntry{
-					{Path: t.TempDir(), Workflow: "W"},
+					{Path: t.TempDir(), MediaType: "M"},
 				},
```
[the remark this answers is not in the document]
solidDoWant: Fixed in 1fcc6cb. All `"M"` literals replaced with `medialib.MovieType`.

internal/watcherconfig/config.go
```diff
@@ -15,36 +19,16 @@ func (CronExpression) JSONSchema() *jsonschema.Schema {
...
+// WatchEntry maps a filesystem path to a media type for dispatch.
 type WatchEntry struct {
 	Path string `yaml:"path" jsonschema:"minLength=1" validate:"min=1"`
-	// Workflow must be one of the values in validWorkflowNames; validated by the workflowname tag.
-	// The validate tag is required for runtime enforcement; JSONSchema() handles schema generation.
-	Workflow WorkflowName `yaml:"workflow" validate:"workflowname"`
+	// MediaType must be one of the values in validMediaTypes; validated by the mediatype tag.
+	// The validate tag is required for runtime enforcement; medialib.MediaType.JSONSchema handles schema generation.
+	MediaType medialib.MediaType `yaml:"media_type" validate:"mediatype"`
 }
```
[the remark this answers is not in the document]
solidDoWant: Fixed in 1fcc6cb. Regenerated `schemas/watcher.schema.json` via `make generate-schema` — `WatchEntry` now has `media_type` (with enum `["movie","show"]`) instead of `workflow`/`WorkflowName`.

pkg/medialib/medialib.go
```diff
@@ -32,11 +52,19 @@ type Episode struct {
...
 type MovieLibrary interface {
 	GetMovieByFilePath(ctx context.Context, path string) (Movie, error)
-	RefreshMovie(ctx context.Context, id int64) error
 }
 
 // EpisodeLibrary provides operations for episode media items.
 type EpisodeLibrary interface {
 	GetEpisodeByFilePath(ctx context.Context, path string) (Episode, error)
-	RefreshSeries(ctx context.Context, seriesID int64) error
+}
```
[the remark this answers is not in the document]
solidDoWant: Fixed in 1fcc6cb. `Episode.SeriesID` comment now references `ArrLibrary.RefreshByFilePath`.

cmd/worker/main.go
```diff
@@ -46,13 +47,30 @@ func run(ctx context.Context) error {
...
 
+	sonarrURL := os.Getenv("SONARR_URL")
+	if sonarrURL == "" {
+		return fmt.Errorf("SONARR_URL is not set")
+	}
+
+	sonarrAPIKey := os.Getenv("SONARR_API_KEY")
+	if sonarrAPIKey == "" {
+		return fmt.Errorf("SONARR_API_KEY is not set")
+	}
```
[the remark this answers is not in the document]
solidDoWant: I think this is reasonable for now. I'm not planning on deploying this to an env where only one is available.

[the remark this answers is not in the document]
solidDoWant: Acknowledged — keeping Sonarr required for now as both are always deployed together.

cmd/watcher/watcher.go
```diff
@@ -110,8 +109,8 @@ func scan(ctx context.Context, cfg *Config, dispatch dispatchFunc) error {
 				return nil
 			}
 
-			if dispatchErr := dispatch(ctx, workflowName, absPath); dispatchErr != nil {
-				errs = append(errs, fmt.Errorf("dispatch workflow %q for %q: %w", workflowName, absPath, dispatchErr))
+			if dispatchErr := dispatch(ctx, absPath, w.MediaType); dispatchErr != nil {
+				errs = append(errs, fmt.Errorf("dispatch workflow for %q: %w", absPath, dispatchErr))
```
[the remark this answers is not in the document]
solidDoWant: Fixed in cde77cd. Error message now includes `w.MediaType` as suggested.

pkg/medialib/radarr/client.go
```diff
@@ -14,8 +14,9 @@ import (
 	"github.com/solidDoWant/media-processor/pkg/medialib"
 )
 
-// Compile-time assertion that *Client implements medialib.MovieLibrary.
+// Compile-time assertions that *Client implements medialib.MovieLibrary and medialib.LibraryClient.
 var _ medialib.MovieLibrary = (*Client)(nil)
+var _ medialib.ArrLibrary = (*Client)(nil)
```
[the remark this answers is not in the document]
solidDoWant: Fixed in cde77cd. Comment now reads "medialib.ArrLibrary".

pkg/medialib/sonarr/client.go
```diff
@@ -13,8 +13,9 @@ import (
 	"github.com/solidDoWant/media-processor/pkg/medialib"
 )
 
-// Compile-time assertion that *Client implements medialib.EpisodeLibrary.
+// Compile-time assertions that *Client implements medialib.EpisodeLibrary and medialib.LibraryClient.
```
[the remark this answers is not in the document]
solidDoWant: Fixed in cde77cd. Comment now reads "medialib.ArrLibrary".

pkg/medialib/medialib.go
```diff
@@ -22,7 +42,7 @@ type Episode struct {
 	// ID is the internal database ID assigned by the backing TV library service.
 	ID int64
 	// SeriesID is the internal database ID of the series this episode belongs to.
-	// Use this with EpisodeLibrary.RefreshSeries to trigger a library rescan.
+	// Use this with ArrLibrary.RefreshByFilePath to trigger a library rescan.
```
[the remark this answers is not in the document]
solidDoWant: Fixed in 2fdf46a. Comment now reads "Used internally by the Sonarr client for series-level refresh."

cmd/worker/main.go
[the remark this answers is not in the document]
solidDoWant: This was already addressed in the earlier thread (r2998430783 / r2998433861): both Radarr and Sonarr are always deployed together, so keeping both env vars required is intentional.

workflows/media/integration_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+	radarrStub := &stubLibraryClient{err: medialib.ErrNotFound}
+	wf := NewMediaWorkflow(client, MediaWorkflowConfig{OutputDir: outputDir}, radarrStub, &stubLibraryClient{}, &webhook.Client{})
+	startMediaWorker(t, client, wf)
+
+	_, err = wf.Run(t.Context(), MediaInput{FilePath: inputPath, MediaType: medialib.MovieType})
+	assert.Error(t, err, "workflow should fail when the movie is not found in Radarr")
+
+	_, statErr := os.Stat(inputPath)
+	assert.NoError(t, statErr, "source file should not be deleted when notify fails")
```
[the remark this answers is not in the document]
solidDoWant: The suggested `assert.Empty` assertion would be incorrect given the current workflow design. Because transcode runs *before* notify (`probe → transcode → notify → cleanup`), when notify fails with `ErrNotFound` the transcoded file has already been written to `outputDir` — so the directory will **not** be empty. Cleanup only runs if notify succeeds, which is the documented trade-off accepted in this PR (see r2998245865 / r2998248349). The assertion to add — if we want to lock in that behavior — would be the opposite: that the output dir is *not* empty. I've left the test as-is since asserting the presence of an intermediate transcoded file is fragile and the trade-off is already documented.

PR #34, opened 2026-03-27
feat(workflows): strip non-English subtitle streams during transcode/transmux

### LINKED ISSUE #33: feat(workflows): strip non-English subtitle streams during transcode/transmux
## Summary

When transcoding or transmuxing a media file, the workflow should discard all
subtitle streams not tagged as English (`eng`). Unlike audio filtering, there is
no safe-fallback: if no subtitle stream is tagged English, all subtitle streams
are removed. This mirrors the general approach introduced for audio streams in #31,
with a stricter fallback policy.

## Acceptance Criteria

- [x] **Given** a media file with multiple subtitle streams where at least one is tagged `eng` and at least one is tagged with a different language code, **when** the workflow processes the file, **then** the output file contains only the subtitle streams tagged `eng`.
- [x] **Given** a media file where every subtitle stream is tagged `eng`, **when** the workflow processes the file, **then** all subtitle streams are present in the output.
- [x] **Given** a media file where no subtitle stream carries a language tag, **when** the workflow processes the file, **then** all subtitle streams are removed from the output.
- [x] **Given** a media file where no subtitle stream is tagged `eng` (e.g. only `jpn` and `fra`), **when** the workflow processes the file, **then** all subtitle streams are removed from the output.

## Technical Context

- Depends on: #31 (merged)
- `ffprobe.StreamInfo` already exposes `Tags map[string]string` (added in #31) and `CodecTypeSubtitle` is already defined — no changes to `pkg/ffprobe` are required.
- `pkg/ffmpeg.TranscodeBuilder.ExcludeStreams()` is stream-type-agnostic and already works for subtitle indices — no changes to `pkg/ffmpeg` are required.
- `workflows/shared.ProbeOutput` needs a `SubtitleStreams []SubtitleStreamInfo` field (parallel to `AudioStreams`); `RunProbe` must populate it by iterating streams with `CodecType == ffprobe.CodecTypeSubtitle`.
- `RunTranscode` must compute non-English subtitle indices and pass them to `ExcludeStreams()`. Unlike `nonEnglishAudioIndices`, there is no safe-fallback: all non-`eng` subtitle streams are always excluded (including untagged ones).
- The ISO 639-2/B tag for English is `eng`.

## Non-Goals

- Language-based filtering of audio or other non-subtitle stream types.
- Configurable language allowlist (English-only for now).
- Re-encoding or converting subtitle formats.

solidDoWant:
Fixes #33

### DIFF
--- a/workflows/shared/probe.go
+++ b/workflows/shared/probe.go
@@ -10,14 +10,8 @@ import (
 	"github.com/solidDoWant/media-processor/pkg/ffprobe"
 )
 
-// AudioStreamInfo holds the input stream index and language tag for one audio stream.
-type AudioStreamInfo struct {
-	Index    int    `json:"index"`
-	Language string `json:"language"`
-}
-
-// SubtitleStreamInfo holds the input stream index and language tag for one subtitle stream.
-type SubtitleStreamInfo struct {
+// StreamInfo holds the input stream index and language tag for one audio or subtitle stream.
+type StreamInfo struct {
 	Index    int    `json:"index"`
 	Language string `json:"language"`
 }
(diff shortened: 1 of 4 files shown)


workflows/shared/probe.go
```diff
@@ -16,6 +16,12 @@ type AudioStreamInfo struct {
 	Language string `json:"language"`
 }
 
+// SubtitleStreamInfo holds the input stream index and language tag for one subtitle stream.
+type SubtitleStreamInfo struct {
+	Index    int    `json:"index"`
+	Language string `json:"language"`
+}
+
```
solidDoWant, line 18: Do you have an opinion on combining this into a single struct?

solidDoWant: Good call — merged them into a single `StreamInfo` struct in 791c06e. Both structs were identical so there was no reason to keep them separate.

PR #36, opened 2026-03-27
feat(workflows): set English audio and subtitle tracks as default in output

### LINKED ISSUE #35: feat(workflows): set English audio and subtitle tracks as default in output
## Summary

After filtering out non-English audio and subtitle streams (issues #31 and #33), the output MKV should mark the first English audio stream and the first English subtitle stream with the `default` disposition flag so media players select them automatically. If no English stream exists for a given type (audio preserved under safe-fallback, or subtitles all stripped), dispositions are left as-is.

This feature is technically feasible: `go-astiav` v0.40.0 exposes `Stream.SetDispositionFlags` and `DispositionFlagDefault`.

Note: currently, output streams have no disposition flags at all (not copied from input). This implementation will also copy input dispositions to output streams as a prerequisite, preserving existing metadata (e.g. `forced`, `hearing_impaired`).

## Acceptance Criteria

- [x] **Given** a media file with at least one English audio stream, **when** the workflow processes the file, **then** the first English audio stream in the output has the `default` disposition flag set and no other audio stream does.
- [x] **Given** a media file with no English audio streams (audio preserved as safe-fallback), **when** the workflow processes the file, **then** audio stream dispositions are unchanged from the input (no `default` flag is forced onto any stream).
- [x] **Given** a media file with at least one English subtitle stream, **when** the workflow processes the file, **then** the first English subtitle stream in the output has the `default` disposition flag set and no other subtitle stream does.
- [x] **Given** a media file where all subtitle streams are stripped (no English subtitles), **when** the workflow processes the file, **then** the output contains no subtitle streams and no disposition logic is applied.
- [x] **Given** a media file with any streams, **when** the workflow processes the file, **then** all non-`default` disposition flags from the input (e.g. `forced`, `hearing_impaired`) are preserved in the output.

## Technical Context

- Depends on: #31 (merged), #33 (merged)
- `go-astiav` v0.40.0 provides `(*Stream).DispositionFlags() DispositionFlags`, `(*Stream).SetDispositionFlags(DispositionFlags)`, `DispositionFlagDefault`, and `DispositionFlagForced`.
- `pkg/ffmpeg.TranscodeBuilder` needs a new `WithDefaultStreams(indices ...int) *TranscodeBuilder` method that records which **input** stream indices should carry the `default` disposition flag in the output.
- `pkg/ffmpeg.setupOutputContext` must be updated to:
  1. Copy the input stream's disposition to the output stream (currently lost — `avcodec_parameters_copy` does not include disposition).
  2. If `defaultStreams` is non-empty for this media type, override `DispositionFlagDefault` so only the specified stream has it set.
- `workflows/shared.RunTranscode` must compute the first English audio index and the first English subtitle index (from `ProbeOutput.AudioStreams` / `SubtitleStreams`) and pass them to `WithDefaultStreams`.
  - "First English" is defined as the first stream in stream order with `Language == "eng"`.
  - If there is no English audio (safe-fallback case from #31), pass no audio default index — do not alter any audio disposition.
  - If there are no English subtitles (all stripped per #33), pass no subtitle default index — nothing to mark.
- No changes to `pkg/ffprobe`, `workflows/media`, or any other package.

## Non-Goals

- Setting the `forced` disposition on subtitle streams (always-on subtitles).
- Configurable language preference.
- Any changes to which streams are included/excluded (that is handled by #31 and #33).

solidDoWant:
Fixes #35

## Summary

- Extends `TranscodeBuilder` with `WithDefaultAudioStream(idx *int)` and `WithDefaultSubtitleStream(idx *int)` methods; a nil argument is a no-op so callers can pass the result of `firstEnglishIndex` directly
- Fixes a pre-existing bug where output stream dispositions were always zeroed (input dispositions not copied); now all streams copy their input disposition before any override is applied
- Adds a `firstEnglishIndex` helper to `workflows/shared` and wires it into `RunTranscode` so the first English audio/subtitle stream is designated as default; no-op when no English stream exists

## Test plan

- [ ] `TestTranscode_DispositionPreserved` — input dispositions copied to output unchanged when no setter is called
- [ ] `TestTranscode_WithDefaultAudioStream` — specified stream gets default flag; other audio streams have it cleared
- [ ] `TestFirstEnglishIndex` — correct stream index selected across all language-tag combinations
- [ ] All existing tests still pass (`make test`)

### DIFF
--- a/pkg/ffmpeg/transcode.go
+++ b/pkg/ffmpeg/transcode.go
@@ -359,15 +359,15 @@ func (t *Transcoder) outputDisposition(inStream *astiav.Stream) astiav.Dispositi
 		defaultStream = t.defaultSubtitleStream
 	}
 
-	if defaultStream != nil {
-		if inStream.Index() == *defaultStream {
-			disp = disp.Add(astiav.DispositionFlagDefault)
-		} else {
-			disp = disp.Del(astiav.DispositionFlagDefault)
-		}
+	if defaultStream == nil {
+		return disp
 	}
 
-	return disp
+	if inStream.Index() == *defaultStream {
+		return disp.Add(astiav.DispositionFlagDefault)
+	}
+
+	return disp.Del(astiav.DispositionFlagDefault)
 }
 
 // readAllPackets is the main decode/encode loop.
--- a/pkg/ffmpeg/transcode.go
+++ b/pkg/ffmpeg/transcode.go
@@ -89,19 +89,19 @@ func (b *TranscodeBuilder) ExcludeStreams(indices ...int) *TranscodeBuilder {
 
 // WithDefaultAudioStream marks the audio stream at the given input stream index
 // as the default audio track in the output. All other audio stream dispositions
-// have their default flag cleared. If not called, audio stream dispositions are
-// copied from the input unchanged.
-func (b *TranscodeBuilder) WithDefaultAudioStream(idx int) *TranscodeBuilder {
-	b.defaultAudioStream = &idx
+// have their default flag cleared. A nil argument is a no-op: audio stream
+// dispositions are copied from the input unchanged.
+func (b *TranscodeBuilder) WithDefaultAudioStream(idx *int) *TranscodeBuilder {
+	b.defaultAudioStream = idx
 	return b
 }
 
 // WithDefaultSubtitleStream marks the subtitle stream at the given input stream
 // index as the default subtitle track in the output. All other subtitle stream
-// dispositions have their default flag cleared. If not called, subtitle stream
-// dispositions are copied from the input unchanged.
-func (b *TranscodeBuilder) WithDefaultSubtitleStream(idx int) *TranscodeBuilder {
-	b.defaultSubtitleStream = &idx
+// dispositions have their default flag cleared. A nil argument is a no-op:
+// subtitle stream dispositions are copied from the input unchanged.
+func (b *TranscodeBuilder) WithDefaultSubtitleStream(idx *int) *TranscodeBuilder {
+	b.defaultSubtitleStream = idx
 	return b
 }
 
--- a/pkg/ffmpeg/transcode.go
+++ b/pkg/ffmpeg/transcode.go
@@ -350,24 +350,23 @@ func (t *Transcoder) setupOutputContext(streams map[int]stream, inputFmt *astiav
 // override configured via WithDefaultAudioStream or WithDefaultSubtitleStream.
 func (t *Transcoder) outputDisposition(inStream *astiav.Stream) astiav.DispositionFlags {
 	disp := inStream.DispositionFlags()
+
+	var defaultStream *int
 	switch inStream.CodecParameters().MediaType() {
 	case astiav.MediaTypeAudio:
-		if t.defaultAudioStream != nil {
-			if inStream.Index() == *t.defaultAudioStream {
-				disp = disp.Add(astiav.DispositionFlagDefault)
-			} else {
-				disp = disp.Del(astiav.DispositionFlagDefault)
-			}
-		}
+		defaultStream = t.defaultAudioStream
 	case astiav.MediaTypeSubtitle:
-		if t.defaultSubtitleStream != nil {
-			if inStream.Index() == *t.defaultSubtitleStream {
-				disp = disp.Add(astiav.DispositionFlagDefault)
-			} else {
-				disp = disp.Del(astiav.DispositionFlagDefault)
-			}
+		defaultStream = t.defaultSubtitleStream
+	}
+
+	if defaultStream != nil {
+		if inStream.Index() == *defaultStream {
+			disp = disp.Add(astiav.DispositionFlagDefault)
+		} else {
+			disp = disp.Del(astiav.DispositionFlagDefault)
 		}
 	}
+
 	return disp
 }
 
--- a/pkg/ffmpeg/transcode.go
+++ b/pkg/ffmpeg/transcode.go
@@ -18,6 +18,8 @@ type TranscodeBuilder struct {
 	progressCh            chan<- Progress
 	startHook             func()
 	excludeStreams        map[int]bool
+	defaultAudioStream    *int // input stream index to mark as default audio; nil = preserve input dispositions
+	defaultSubtitleStream *int // input stream index to mark as default subtitle; nil = preserve input dispositions
 }
 
 // NewTranscode returns a builder for a transcode job from inputPath to outputPath.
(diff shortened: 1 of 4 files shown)


workflows/shared/transcode.go
```diff
@@ -62,6 +62,18 @@ func nonEnglishSubtitleIndices(streams []StreamInfo) []int {
 	return exclude
 }
 
+// firstEnglishIndex returns the input stream Index of the first StreamInfo
+// element tagged "eng". The second return value is false when no English stream
+// is found.
+func firstEnglishIndex(streams []StreamInfo) (int, bool) {
```
solidDoWant: Would it be simpler if this just returned `*int` instead of `int, bool`?

solidDoWant: Good call — updated to return `*int` with `nil` as the not-found sentinel. Pushed in the follow-up commit.

workflows/shared/transcode.go
```diff
@@ -62,6 +62,18 @@ func nonEnglishSubtitleIndices(streams []StreamInfo) []int {
 	return exclude
 }
 
+// firstEnglishIndex returns a pointer to the input stream Index of the first
+// StreamInfo element tagged "eng", or nil when no English stream is found.
+func firstEnglishIndex(streams []StreamInfo) *int {
+	for _, s := range streams {
+		if s.Language == "eng" {
+			idx := s.Index
+			return &idx
```
solidDoWant:
```suggestion
			return &s.Index
```

workflows/shared/transcode.go
```diff
@@ -77,12 +89,18 @@ func RunTranscode(ctx context.Context, filePath string, probe ProbeOutput, outpu
...
-		ExcludeStreams(excludeIndices...).
-		Build().
-		Run(ctx); err != nil {
+		ExcludeStreams(excludeIndices...)
+	if idx := firstEnglishIndex(probe.AudioStreams); idx != nil {
+		builder = builder.WithDefaultAudioStream(*idx)
+	}
+	if idx := firstEnglishIndex(probe.SubtitleStreams); idx != nil {
+		builder = builder.WithDefaultSubtitleStream(*idx)
+	}
```
solidDoWant: Are these conditionals actually needed now? Can the `WithDefault` funcs just be chained normally?

workflows/shared/transcode.go
solidDoWant: Applied — simplified to `return &s.Index` directly.

workflows/shared/transcode.go
solidDoWant: Yes — changed `WithDefaultAudioStream` and `WithDefaultSubtitleStream` to accept `*int` with nil as a no-op, so `firstEnglishIndex` can be passed directly and the conditionals are gone. `RunTranscode` is now a single fluent chain.

pkg/ffmpeg/transcode.go
```diff
@@ -322,6 +345,32 @@ func (t *Transcoder) setupOutputContext(streams map[int]stream, inputFmt *astiav
...
+		}
+	case astiav.MediaTypeSubtitle:
+		if t.defaultSubtitleStream != nil {
+			if inStream.Index() == *t.defaultSubtitleStream {
+				disp = disp.Add(astiav.DispositionFlagDefault)
+			} else {
+				disp = disp.Del(astiav.DispositionFlagDefault)
+			}
+		}
+	}
```
solidDoWant, line 368: I think that this could be sipmlified if instead of duplicating the add/del logic here, a "defaultStream" variable was set via `switch`, and the conditional `if defaultStream != nil` logic was outside of the switch.

solidDoWant: Done — the switch now just sets a `defaultStream *int` local, and the single add/del block runs outside it.

pkg/ffmpeg/transcode.go
```diff
@@ -85,6 +87,24 @@ func (b *TranscodeBuilder) ExcludeStreams(indices ...int) *TranscodeBuilder {
...
+}
+
+// WithDefaultSubtitleStream marks the subtitle stream at the given input stream
+// index as the default subtitle track in the output. All other subtitle stream
+// dispositions have their default flag cleared. A nil argument is a no-op:
+// subtitle stream dispositions are copied from the input unchanged.
+func (b *TranscodeBuilder) WithDefaultSubtitleStream(idx *int) *TranscodeBuilder {
+	b.defaultSubtitleStream = idx
+	return b
+}
```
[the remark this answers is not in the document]
solidDoWant: pr desc

workflows/shared/transcode.go
```diff
@@ -62,6 +62,17 @@ func nonEnglishSubtitleIndices(streams []StreamInfo) []int {
 	return exclude
 }
 
+// firstEnglishIndex returns a pointer to the input stream Index of the first
+// StreamInfo element tagged "eng", or nil when no English stream is found.
+func firstEnglishIndex(streams []StreamInfo) *int {
+	for _, s := range streams {
+		if s.Language == "eng" {
+			return &s.Index
```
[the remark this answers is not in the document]
solidDoWant: Does this make a functional difference?

However, `s` should be changed  to `stream`.

pkg/ffmpeg/transcode_test.go
```diff
@@ -236,6 +255,70 @@ func TestTranscode_ExcludeStreams(t *testing.T) {
...
+// TestTranscode_WithDefaultAudioStream verifies that WithDefaultAudioStream
+// marks the specified audio stream as default and clears the flag from all
+// other audio streams.
+func TestTranscode_WithDefaultAudioStream(t *testing.T) {
+	// Test fixture (video.mp4): stream 0 = video, stream 1 = audio (both default=1 in input).
+	tests := []struct {
+		name        string
+		audioIdx    int
+		wantDefault bool
+	}{
```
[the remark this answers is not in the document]
solidDoWant: File an issue for this

pkg/ffmpeg/transcode.go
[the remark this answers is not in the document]
solidDoWant: PR description updated to reflect `*int` signatures.

pkg/ffmpeg/transcode.go
```diff
@@ -322,6 +345,31 @@ func (t *Transcoder) setupOutputContext(streams map[int]stream, inputFmt *astiav
...
+	if defaultStream == nil {
+		return disp
+	}
+
+	if inStream.Index() == *defaultStream {
+		return disp.Add(astiav.DispositionFlagDefault)
+	}
+
+	return disp.Del(astiav.DispositionFlagDefault)
+}
```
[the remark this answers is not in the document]
solidDoWant: Filed #38 to track adding a subtitle fixture and covering this code path.

workflows/shared/transcode.go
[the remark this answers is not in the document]
solidDoWant: Renamed to `stream` — pushed in the follow-up commit. There is no functional difference (`&s.Index` and `&stream.Index` both take the address of the loop-copy's field and escape to the heap when returned), but `stream` is clearer.

pkg/ffmpeg/transcode_test.go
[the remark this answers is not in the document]
solidDoWant: Filed #37 to track adding a multi-audio fixture and covering this case.

PR #47, opened 2026-03-28
feat(ffmpeg): configurable hardware device path for transcoding

### LINKED ISSUE #46: feat(ffmpeg): configurable hardware device path for transcoding
## Summary

In deployments with multiple transcoding devices (e.g. multiple VAAPI render nodes or CUDA GPUs), ffmpeg currently auto-selects the device because the device path passed to `CreateHardwareDeviceContext` is always `""`. This change adds an optional `HARDWARE_DEVICE_PATH` environment variable read by `cmd/worker` and threaded through to the `TranscodeBuilder`, enabling operators to pin transcoding to a specific device.

## Acceptance Criteria

- [x] **Given** a `TranscodeBuilder` with a device path set via `WithHardwareDevice`, **when** the transcoder initialises the hardware device context, **then** the specified path is passed to `CreateHardwareDeviceContext` for both the decoder and encoder device contexts.
- [x] **Given** a `TranscodeBuilder` with no device path set, **when** the transcoder initialises the hardware device context, **then** an empty string is passed to `CreateHardwareDeviceContext` (preserving existing auto-select behaviour).
- [x] **Given** a `TranscodeBuilder` configured with `WithHardwareDevice("")`, **when** `Build` is called, **then** the builder behaves identically to one with no device path set (empty string is a no-op).
- [x] **Given** `HARDWARE_DEVICE_PATH` is set in the worker's environment, **when** `cmd/worker` starts, **then** the value is passed to the media workflow and used as the hardware device path for all transcode jobs.
- [x] **Given** `HARDWARE_DEVICE_PATH` is not set, **when** `cmd/worker` starts, **then** the worker starts normally and hardware device selection falls back to the libav default (auto-select).

## Technical Context

- Depends on: none
- `astiav.CreateHardwareDeviceContext(deviceType, devicePath string, ...)` already accepts a device path; both call sites in `pkg/ffmpeg/stream_video.go` (decoder setup ~line 87, encoder setup ~line 191) currently pass `""`.
- Add `WithHardwareDevice(path string) *TranscodeBuilder` to `pkg/ffmpeg`; store the path on `TranscodeBuilder` and thread it through to both `CreateHardwareDeviceContext` calls.
- `cmd/worker` reads the env var and passes it into `MediaWorkflowConfig`, which threads it through to wherever `NewTranscode` is called. Follow the same optional env var pattern as `RADARR_LOCAL_PATH_PREFIX` (read with `os.Getenv`, no error if absent).
- Typical device paths: `/dev/dri/renderD128` or `/dev/dri/renderD129` for VAAPI; `"0"` or `"1"` (device index) for CUDA; a DRI path for QSV.
- When the hw device context is shared between decoder and encoder (zero-copy path), only one `CreateHardwareDeviceContext` call is made — the device path must still be used there.

## Non-Goals

- Device enumeration or discovery.
- Selecting different devices for decode vs. encode within a single transcode job.
- Per-workflow or per-job device override (env var applies globally to the worker process).

solidDoWant:
Fixes #46

### DIFF
--- a/pkg/ffmpeg/transcode.go
+++ b/pkg/ffmpeg/transcode.go
@@ -16,6 +16,7 @@ type TranscodeBuilder struct {
 	audioCodec            Codec
 	container             Container
 	hwAccel               HWAccel
+	hardwareDevicePath    string // device path passed to CreateHardwareDeviceContext; "" = auto-select
 	progressCh            chan<- Progress
 	startHook             func()
 	excludeStreams        map[int]bool
@@ -62,6 +63,19 @@ func (b *TranscodeBuilder) ToContainer(c Container) *TranscodeBuilder {
 // HardwareAccel sets the hardware acceleration mode.
 func (b *TranscodeBuilder) HardwareAccel(h HWAccel) *TranscodeBuilder {
 	b.hwAccel = h
+	return b
+}
+
+// WithHardwareDevice sets the device path passed to CreateHardwareDeviceContext
+// for both the decoder and encoder hardware device contexts. Typical values are
+// "/dev/dri/renderD128" for VAAPI/QSV or "0"/"1" for CUDA device indices.
+// An empty string is a no-op: the device path remains unset and libav will
+// auto-select the hardware device (preserving existing behaviour).
+func (b *TranscodeBuilder) WithHardwareDevice(path string) *TranscodeBuilder {
+	if path == "" {
+		return b
+	}
+	b.hardwareDevicePath = path
 	return b
 }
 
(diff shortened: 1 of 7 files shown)


pkg/ffmpeg/transcode.go
```diff
@@ -65,6 +66,19 @@ func (b *TranscodeBuilder) HardwareAccel(h HWAccel) *TranscodeBuilder {
...
 
+// WithHardwareDevice sets the device path passed to CreateHardwareDeviceContext
+// for both the decoder and encoder hardware device contexts. Typical values are
+// "/dev/dri/renderD128" for VAAPI/QSV or "0"/"1" for CUDA device indices.
+// An empty string is a no-op: the device path remains unset and libav will
+// auto-select the hardware device (preserving existing behaviour).
+func (b *TranscodeBuilder) WithHardwareDevice(path string) *TranscodeBuilder {
+	if path == "" {
+		return b
+	}
```
solidDoWant: Is this actually necessary?

solidDoWant: You're right, it's not — since `hardwareDevicePath` defaults to `""` (Go's zero value), assigning `""` is already a no-op. Removed in 85f1516.

PR #58, opened 2026-03-29
feat(worker): add metrics infrastructure with Prometheus pull endpoint and OTLP push export

### LINKED ISSUE #54: feat(worker): add metrics infrastructure with Prometheus pull endpoint and OTLP push export
## Summary

Add a metrics infrastructure layer to the worker: a pull-based Prometheus `/metrics` HTTP endpoint (controlled by `METRICS_ADDR`) and a push-based OTLP export (controlled by the standard `OTEL_EXPORTER_OTLP_ENDPOINT` env var). Both are optional and independent. On graceful shutdown, buffered OTLP metrics are flushed before exit.

## Technical Context

Parent issue: #50

### Implementation notes

Create a new `pkg/metrics` package (or inline in `cmd/worker`) that:

1. **Prometheus pull endpoint** — when `METRICS_ADDR` is set, start an HTTP server at that address serving `/metrics` in Prometheus text format via `prometheus/client_golang`. When unset, skip the server entirely.

2. **OTLP push export** — when `OTEL_EXPORTER_OTLP_ENDPOINT` is set, configure an OTel SDK `MeterProvider` with an OTLP metric exporter (use the standard OTel Go SDK: `go.opentelemetry.io/otel`). When unset, no OTLP connection is attempted.

3. **Both simultaneously** — if both env vars are set, both exporters are active.

4. **Graceful flush** — the `MeterProvider` must be shut down (calling `Shutdown`) before the worker process exits when `OTEL_EXPORTER_OTLP_ENDPOINT` is configured. Wire this into the existing SIGTERM/SIGINT shutdown in `cmd/worker/main.go`.

Expose a `MetricsProvider` or similar type with a `MustRegister` / `NewHistogram` / `NewCounter` interface that downstream workflow instrumentation (sub-issue #54) can use.

### Required Go module dependencies

- `go.opentelemetry.io/otel`
- `go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetricgrpc` (or HTTP variant)
- `go.opentelemetry.io/otel/sdk/metric`
- `github.com/prometheus/client_golang`

## Acceptance Criteria

- [x] **Given** `METRICS_ADDR` is set to a valid address, **when** the worker starts, **then** a `/metrics` endpoint is available at that address serving Prometheus-format metrics.
- [x] **Given** `METRICS_ADDR` is unset or empty, **when** the worker starts, **then** no metrics HTTP server is started.
- [x] **Given** `OTEL_EXPORTER_OTLP_ENDPOINT` is set, **when** the worker starts, **then** an OTLP metric exporter is initialised targeting that endpoint.
- [x] **Given** `OTEL_EXPORTER_OTLP_ENDPOINT` is unset or empty, **when** the worker shuts down, **then** no OTLP export is attempted.
- [x] **Given** both `METRICS_ADDR` and `OTEL_EXPORTER_OTLP_ENDPOINT` are set, **when** the worker is running, **then** both exporters are active simultaneously.
- [x] **Given** the worker is shut down gracefully (SIGTERM/SIGINT) and `OTEL_EXPORTER_OTLP_ENDPOINT` is set, **then** the OTel `MeterProvider` is shut down (flushing buffered metrics) before the process exits.

## Non-Goals

- Defining or recording specific workflow metrics (sub-issue #54).
- Watcher metrics.
- Tracing / span instrumentation.

solidDoWant:
Fixes #54

### DIFF
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -30,10 +30,14 @@ func main() {
 }
 
 func run(ctx context.Context) error {
-	metricsProvider, err := metrics.New(ctx, metrics.Config{
-		MetricsAddr:  os.Getenv("METRICS_ADDR"),
-		OTLPEndpoint: os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT"),
-	})
+	var metricsOpts []metrics.Option
+	if addr := os.Getenv("METRICS_ADDR"); addr != "" {
+		metricsOpts = append(metricsOpts, metrics.WithMetricsAddr(addr))
+	}
+	if endpoint := os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT"); endpoint != "" {
+		metricsOpts = append(metricsOpts, metrics.WithOTLPEndpoint(endpoint))
+	}
+	metricsProvider, err := metrics.New(ctx, metricsOpts...)
 	if err != nil {
 		return fmt.Errorf("init metrics: %w", err)
 	}
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"os/signal"
 	"syscall"
+	"time"
 
 	hatchet "github.com/hatchet-dev/hatchet/sdks/go"
 
@@ -34,7 +35,9 @@ func run(ctx context.Context) error {
 		return fmt.Errorf("init metrics: %w", err)
 	}
 	defer func() {
-		if err := metricsProvider.Shutdown(context.Background()); err != nil {
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+		defer cancel()
+		if err := metricsProvider.Shutdown(ctx); err != nil {
 			log.Printf("metrics shutdown error: %v", err)
 		}
 	}()
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/solidDoWant/media-processor/pkg/medialib/radarr"
 	"github.com/solidDoWant/media-processor/pkg/medialib/sonarr"
+	"github.com/solidDoWant/media-processor/pkg/metrics"
 	"github.com/solidDoWant/media-processor/pkg/webhook"
 	"github.com/solidDoWant/media-processor/workflows"
 	"github.com/solidDoWant/media-processor/workflows/media"
@@ -28,6 +29,16 @@ func main() {
 }
 
 func run(ctx context.Context) error {
+	metricsProvider, err := metrics.New(ctx)
+	if err != nil {
+		return fmt.Errorf("init metrics: %w", err)
+	}
+	defer func() {
+		if err := metricsProvider.Shutdown(context.Background()); err != nil {
+			log.Printf("metrics shutdown error: %v", err)
+		}
+	}()
+
 	if os.Getenv("HATCHET_CLIENT_TOKEN") == "" {
 		return fmt.Errorf("HATCHET_CLIENT_TOKEN is not set")
 	}
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -30,7 +30,10 @@ func main() {
 }
 
 func run(ctx context.Context) error {
-	metricsProvider, err := metrics.New(ctx)
+	metricsProvider, err := metrics.New(ctx, metrics.Config{
+		MetricsAddr:  os.Getenv("METRICS_ADDR"),
+		OTLPEndpoint: os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT"),
+	})
 	if err != nil {
 		return fmt.Errorf("init metrics: %w", err)
 	}
--- a/go.mod
+++ b/go.mod
@@ -9,8 +9,13 @@ require (
 	github.com/go-playground/validator/v10 v10.30.1
 	github.com/hatchet-dev/hatchet v0.83.1
 	github.com/invopop/jsonschema v0.13.0
+	github.com/prometheus/client_golang v1.23.2
 	github.com/robfig/cron/v3 v3.0.1
 	github.com/stretchr/testify v1.11.1
+	go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetricgrpc v1.42.0
+	go.opentelemetry.io/otel/exporters/prometheus v0.64.0
+	go.opentelemetry.io/otel/metric v1.42.0
+	go.opentelemetry.io/otel/sdk/metric v1.42.0
 	golift.io/starr v1.3.1
 	gopkg.in/yaml.v3 v3.0.1
 )
@@ -69,7 +74,6 @@ require (
 	github.com/perimeterx/marshmallow v1.1.5 // indirect
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
-	github.com/prometheus/client_golang v1.23.2 // indirect
 	github.com/prometheus/client_model v0.6.2 // indirect
 	github.com/prometheus/common v0.67.5 // indirect
 	github.com/prometheus/otlptranslator v1.0.0 // indirect
@@ -88,13 +92,9 @@ require (
 	github.com/woodsbury/decimal128 v1.4.0 // indirect
 	go.opentelemetry.io/auto/sdk v1.2.1 // indirect
 	go.opentelemetry.io/otel v1.42.0 // indirect
-	go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetricgrpc v1.42.0 // indirect
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace v1.42.0 // indirect
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc v1.42.0 // indirect
-	go.opentelemetry.io/otel/exporters/prometheus v0.64.0 // indirect
-	go.opentelemetry.io/otel/metric v1.42.0 // indirect
 	go.opentelemetry.io/otel/sdk v1.42.0 // indirect
-	go.opentelemetry.io/otel/sdk/metric v1.42.0 // indirect
 	go.opentelemetry.io/otel/trace v1.42.0 // indirect
 	go.opentelemetry.io/proto/otlp v1.10.0 // indirect
 	go.yaml.in/yaml/v2 v2.4.3 // indirect
--- a/go.mod
+++ b/go.mod
@@ -22,6 +22,7 @@ require (
 	github.com/apapsch/go-jsonmerge/v2 v2.0.0 // indirect
 	github.com/asticode/go-astikit v0.59.0 // indirect
 	github.com/bahlo/generic-list-go v0.2.0 // indirect
+	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/buger/jsonparser v1.1.1 // indirect
 	github.com/cenkalti/backoff/v5 v5.0.3 // indirect
 	github.com/cespare/xxhash/v2 v2.3.0 // indirect
@@ -60,6 +61,7 @@ require (
 	github.com/mattn/go-colorable v0.1.14 // indirect
 	github.com/mattn/go-isatty v0.0.20 // indirect
 	github.com/mohae/deepcopy v0.0.0-20170929034955-c48cc78d4826 // indirect
+	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
 	github.com/oapi-codegen/runtime v1.3.1 // indirect
 	github.com/oasdiff/yaml v0.0.1 // indirect
 	github.com/oasdiff/yaml3 v0.0.1 // indirect
@@ -67,6 +69,11 @@ require (
 	github.com/perimeterx/marshmallow v1.1.5 // indirect
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
+	github.com/prometheus/client_golang v1.23.2 // indirect
+	github.com/prometheus/client_model v0.6.2 // indirect
+	github.com/prometheus/common v0.67.5 // indirect
+	github.com/prometheus/otlptranslator v1.0.0 // indirect
+	github.com/prometheus/procfs v0.19.2 // indirect
 	github.com/rogpeppe/go-internal v1.14.1 // indirect
 	github.com/rs/zerolog v1.34.0 // indirect
 	github.com/sagikazarmark/locafero v0.12.0 // indirect
@@ -81,12 +88,16 @@ require (
 	github.com/woodsbury/decimal128 v1.4.0 // indirect
 	go.opentelemetry.io/auto/sdk v1.2.1 // indirect
 	go.opentelemetry.io/otel v1.42.0 // indirect
+	go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetricgrpc v1.42.0 // indirect
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace v1.42.0 // indirect
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc v1.42.0 // indirect
+	go.opentelemetry.io/otel/exporters/prometheus v0.64.0 // indirect
 	go.opentelemetry.io/otel/metric v1.42.0 // indirect
 	go.opentelemetry.io/otel/sdk v1.42.0 // indirect
+	go.opentelemetry.io/otel/sdk/metric v1.42.0 // indirect
 	go.opentelemetry.io/otel/trace v1.42.0 // indirect
 	go.opentelemetry.io/proto/otlp v1.10.0 // indirect
+	go.yaml.in/yaml/v2 v2.4.3 // indirect
 	go.yaml.in/yaml/v3 v3.0.4 // indirect
 	golang.org/x/crypto v0.49.0 // indirect
 	golang.org/x/exp v0.0.0-20260312153236-7ab1446f8b90 // indirect
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -18,45 +18,53 @@ import (
 	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
 )
 
+// Config holds the configuration for the metrics Provider.
+type Config struct {
+	// MetricsAddr is the TCP address for the Prometheus /metrics HTTP server (e.g. ":9090").
+	// When empty, no HTTP server is started.
+	MetricsAddr string
+
+	// OTLPEndpoint is the OTLP gRPC endpoint URL (e.g. "http://otel-collector:4317").
+	// When empty, no OTLP exporter is created.
+	OTLPEndpoint string
+}
+
 // Provider manages metrics exporters: a Prometheus pull endpoint and/or an OTLP push exporter.
-// Both are optional and independently controlled by environment variables:
-//   - METRICS_ADDR: if set, starts a Prometheus /metrics HTTP server at that address.
-//   - OTEL_EXPORTER_OTLP_ENDPOINT: if set, initialises an OTLP metric exporter.
+// Both are optional and independently controlled by the Config fields passed to New.
 type Provider struct {
 	meterProvider otelmetric.MeterProvider
 	shutdown      func(context.Context) error
 }
 
-type config struct {
+type internalConfig struct {
 	prometheusListener net.Listener
 }
 
 // Option configures Provider construction.
-type Option func(*config)
+type Option func(*internalConfig)
 
 // WithPrometheusListener supplies a pre-bound listener for the Prometheus HTTP server.
-// METRICS_ADDR must still be set to enable the server; the provided listener is used
-// instead of binding a new one. The caller is responsible for ensuring the listener
-// address matches METRICS_ADDR. Primarily useful in tests to eliminate TOCTOU races.
+// Config.MetricsAddr must be non-empty to enable the server; the provided listener is
+// used instead of binding a new one. The caller is responsible for ensuring the listener
+// address matches Config.MetricsAddr. Primarily useful in tests to eliminate TOCTOU races.
 func WithPrometheusListener(l net.Listener) Option {
-	return func(c *config) {
+	return func(c *internalConfig) {
 		c.prometheusListener = l
 	}
 }
 
-// New creates a Provider based on environment variables. If neither METRICS_ADDR nor
-// OTEL_EXPORTER_OTLP_ENDPOINT is set, a no-op MeterProvider is returned.
-func New(ctx context.Context, opts ...Option) (*Provider, error) {
-	cfg := &config{}
+// New creates a Provider from cfg. If neither MetricsAddr nor OTLPEndpoint is set,
+// a no-op MeterProvider is returned.
+func New(ctx context.Context, cfg Config, opts ...Option) (*Provider, error) {
+	icfg := &internalConfig{}
 	for _, o := range opts {
-		o(cfg)
+		o(icfg)
 	}
 
 	var readers []sdkmetric.Reader
 	var shutdownFuncs []func(context.Context) error
 
-	metricsAddr := os.Getenv("METRICS_ADDR")
-	if metricsAddr != "" {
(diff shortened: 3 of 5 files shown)


cmd/worker/main.go
```diff
@@ -28,6 +29,16 @@ func main() {
 }
 
 func run(ctx context.Context) error {
+	metricsProvider, err := metrics.New(ctx)
+	if err != nil {
+		return fmt.Errorf("init metrics: %w", err)
+	}
+	defer func() {
+		if err := metricsProvider.Shutdown(context.Background()); err != nil {
```
solidDoWant: The shutdown context should be time limited to something reasonable

go.mod
solidDoWant, line 1: run go mod tidy

cmd/worker/main.go
solidDoWant: Fixed — the shutdown context is now bounded to 10 seconds.

go.mod
solidDoWant: Done — ran `go mod tidy`.

go.mod
```diff
@@ -60,13 +66,18 @@ require (
...
 	github.com/mattn/go-isatty v0.0.20 // indirect
 	github.com/mohae/deepcopy v0.0.0-20170929034955-c48cc78d4826 // indirect
+	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
 	github.com/oapi-codegen/runtime v1.3.1 // indirect
 	github.com/oasdiff/yaml v0.0.1 // indirect
 	github.com/oasdiff/yaml3 v0.0.1 // indirect
 	github.com/pelletier/go-toml/v2 v2.3.0 // indirect
 	github.com/perimeterx/marshmallow v1.1.5 // indirect
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
```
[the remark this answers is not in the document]
solidDoWant: run go mod tidy

go.mod
```diff
@@ -83,10 +94,10 @@ require (
 	go.opentelemetry.io/otel v1.42.0 // indirect
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace v1.42.0 // indirect
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc v1.42.0 // indirect
-	go.opentelemetry.io/otel/metric v1.42.0 // indirect
 	go.opentelemetry.io/otel/sdk v1.42.0 // indirect
```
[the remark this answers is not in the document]
solidDoWant: run go mod tidy

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	otlpEndpoint := os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT")
+	if otlpEndpoint != "" {
+		otlpExporter, err := otlpmetricgrpc.New(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("create OTLP metric exporter: %w", err)
+		}
+		readers = append(readers, sdkmetric.NewPeriodicReader(otlpExporter))
+	}
```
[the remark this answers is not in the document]
solidDoWant: Fixed — `otlpmetricgrpc.New` now receives `otlpmetricgrpc.WithEndpointURL(otlpEndpoint)` explicitly, so the configuration path is unambiguous.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,97 @@
...
+)
+
+// Provider manages metrics exporters: a Prometheus pull endpoint and/or an OTLP push exporter.
+// Both are optional and independently controlled by environment variables:
+//   - METRICS_ADDR: if set, starts a Prometheus /metrics HTTP server at that address.
+//   - OTEL_EXPORTER_OTLP_ENDPOINT: if set, initialises an OTLP metric exporter.
+type Provider struct {
+	meterProvider otelmetric.MeterProvider
+	shutdown      func(context.Context) error
+}
```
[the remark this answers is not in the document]
solidDoWant: The `MustRegister`/`NewHistogram`/`NewCounter` wrapper is explicitly listed under Non-Goals ("Defining or recording specific workflow metrics"). The `MeterProvider()` interface is the intended hook for downstream instrumentation — callers create instruments directly from it using the standard OTel API.

cmd/worker/main.go
```diff
@@ -28,6 +30,18 @@ func main() {
...
 
 func run(ctx context.Context) error {
+	metricsProvider, err := metrics.New(ctx)
+	if err != nil {
+		return fmt.Errorf("init metrics: %w", err)
+	}
+	defer func() {
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+		defer cancel()
+		if err := metricsProvider.Shutdown(ctx); err != nil {
```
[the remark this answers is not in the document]
solidDoWant: Already addressed in the previous commit (bounded to 10 seconds). Using `context.WithTimeout(ctx, ...)` with the worker shutdown `ctx` would not work here — `ctx` is already cancelled by the time the deferred function runs (after `worker.StartBlocking` returns), so the shutdown context would be immediately cancelled too. `context.Background()` with a fixed timeout is the right approach.

go.mod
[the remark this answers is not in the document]
solidDoWant: Fixed in the previous commit — `go mod tidy` was run and the direct deps are now correctly marked without `// indirect`.

go.mod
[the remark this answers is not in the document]
solidDoWant: Fixed — `go mod tidy` was run in the previous commit.

pkg/metrics/metrics_test.go
```diff
@@ -0,0 +1,168 @@
...
+)
+
+// freeAddr returns a random free TCP address on localhost.
+func freeAddr(t *testing.T) string {
+	t.Helper()
+	l, err := net.Listen("tcp", "127.0.0.1:0")
+	require.NoError(t, err)
+	addr := l.Addr().String()
+	require.NoError(t, l.Close())
+	return addr
```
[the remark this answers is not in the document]
solidDoWant: Fixed — added `WithPrometheusListener(l net.Listener)` option so tests can pass a pre-bound listener directly to `New`, eliminating the TOCTOU window entirely. `New` still calls `net.Listen` when no listener is provided (production path).

pkg/metrics/metrics.go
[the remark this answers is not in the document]
solidDoWant: Fixed — if `otlpmetricgrpc.New` returns an error, the code now calls `fn(ctx)` for each already-registered shutdown func (including `srv.Shutdown`) before returning the error.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,97 @@
...
+		readers = append(readers, promReader)
+
+		mux := http.NewServeMux()
+		mux.Handle("/metrics", promhttp.HandlerFor(reg, promhttp.HandlerOpts{}))
+		srv := &http.Server{
+			Addr:    metricsAddr,
+			Handler: mux,
+		}
+		go func() { _ = srv.ListenAndServe() }()
+		shutdownFuncs = append(shutdownFuncs, srv.Shutdown)
```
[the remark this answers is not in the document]
solidDoWant: Fixed — replaced `srv.ListenAndServe()` with `net.Listen` + `srv.Serve(l)`. The `net.Listen` call happens synchronously in `New`, so any bind failure (port in use, invalid address, etc.) is returned as an error immediately.

pkg/metrics/metrics_test.go
```diff
@@ -0,0 +1,149 @@
...
+	url := "http://" + l.Addr().String() + "/metrics"
+	resp, err := http.Get(url) //nolint:noctx
+	require.NoError(t, err)
+	t.Cleanup(func() { require.NoError(t, resp.Body.Close()) })
+
+	assert.Equal(t, http.StatusOK, resp.StatusCode)
+	body, err := io.ReadAll(resp.Body)
+	require.NoError(t, err)
+	// Prometheus text format always contains comment lines starting with '#'.
+	assert.True(t, strings.Contains(string(body), "#"), "response body should contain Prometheus-format metrics")
```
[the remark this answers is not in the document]
solidDoWant: Fixed — replaced the `#` body check with a `Content-Type` header assertion (`text/plain` prefix), which is robust regardless of whether the registry is empty.

pkg/metrics/metrics_test.go
```diff
@@ -0,0 +1,149 @@
...
+	t.Setenv("METRICS_ADDR", l.Addr().String())
+	t.Setenv("OTEL_EXPORTER_OTLP_ENDPOINT", "")
+
+	p, err := metrics.New(t.Context(), metrics.WithPrometheusListener(l))
+	require.NoError(t, err)
+	defer p.Shutdown(context.Background()) //nolint:errcheck
+
+	// The server is already listening on l; no poll needed.
+	url := "http://" + l.Addr().String() + "/metrics"
+	resp, err := http.Get(url) //nolint:noctx
```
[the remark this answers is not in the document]
solidDoWant: Fixed — all `http.Get` calls in tests now use `&http.Client{Timeout: 5 * time.Second}`.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	otlpEndpoint := os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT")
+	if otlpEndpoint != "" {
+		otlpExporter, err := otlpmetricgrpc.New(ctx, otlpmetricgrpc.WithEndpointURL(otlpEndpoint))
+		if err != nil {
+			// Shut down any already-started servers before returning.
+			for _, fn := range shutdownFuncs {
+				_ = fn(ctx)
+			}
```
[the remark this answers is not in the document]
solidDoWant: Fixed — error-path cleanup now uses `context.WithTimeout(context.Background(), 5*time.Second)` so the HTTP server shutdown runs regardless of whether the incoming `ctx` is already cancelled.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	mp := sdkmetric.NewMeterProvider(opts2...)
+	shutdownFuncs = append(shutdownFuncs, mp.Shutdown)
+
+	return &Provider{
+		meterProvider: mp,
+		shutdown: func(ctx context.Context) error {
+			var errs []error
+			for _, fn := range shutdownFuncs {
+				errs = append(errs, fn(ctx))
```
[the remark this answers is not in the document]
solidDoWant: Fixed — shutdown now iterates in reverse (LIFO) order, so `mp.Shutdown` (OTLP flush) runs before `srv.Shutdown`, giving the flush the full context deadline.

pkg/metrics/metrics_test.go
```diff
@@ -0,0 +1,151 @@
...
+	url := "http://" + l.Addr().String() + "/metrics"
+	client := &http.Client{Timeout: 5 * time.Second}
+	resp, err := client.Get(url) //nolint:noctx
+	require.NoError(t, err)
+	t.Cleanup(func() { require.NoError(t, resp.Body.Close()) })
+
+	assert.Equal(t, http.StatusOK, resp.StatusCode)
+	// Verify the response Content-Type is Prometheus text exposition format.
+	contentType := resp.Header.Get("Content-Type")
+	assert.True(t, len(contentType) > 0 && contentType[:10] == "text/plain", "metrics endpoint should return Prometheus text exposition format, got: %s", contentType)
```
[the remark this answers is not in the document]
solidDoWant: Fixed — replaced `contentType[:10]` with `strings.HasPrefix(contentType, "text/plain")`, which is panic-safe for any header value.

pkg/metrics/metrics_test.go
```diff
@@ -0,0 +1,151 @@
...
+	promListener := bindListener(t)
+	otlpListener := bindListener(t)
+
+	t.Setenv("METRICS_ADDR", promListener.Addr().String())
+	t.Setenv("OTEL_EXPORTER_OTLP_ENDPOINT", "http://"+otlpListener.Addr().String())
+
+	p, err := metrics.New(t.Context(), metrics.WithPrometheusListener(promListener))
+	require.NoError(t, err)
+	defer p.Shutdown(context.Background()) //nolint:errcheck
+
```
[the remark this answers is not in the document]
solidDoWant: Fixed — `TestBothExporters_Active` now uses `t.Cleanup` with a 5-second bounded context for shutdown instead of `defer p.Shutdown(context.Background())`.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,135 @@
...
+			l, err = net.Listen("tcp", metricsAddr)
+			if err != nil {
+				return nil, fmt.Errorf("listen on metrics addr %s: %w", metricsAddr, err)
+			}
+		}
+
+		mux := http.NewServeMux()
+		mux.Handle("/metrics", promhttp.HandlerFor(reg, promhttp.HandlerOpts{}))
+		srv := &http.Server{Handler: mux}
+		go func() { _ = srv.Serve(l) }()
```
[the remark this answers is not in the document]
solidDoWant: Fixed — the goroutine now logs non-`http.ErrServerClosed` errors to stderr so unexpected failures are visible.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+		readers = append(readers, promReader)
+
+		l := cfg.prometheusListener
+		if l == nil {
+			l, err = net.Listen("tcp", metricsAddr)
+			if err != nil {
+				return nil, fmt.Errorf("listen on metrics addr %s: %w", metricsAddr, err)
+			}
+		}
```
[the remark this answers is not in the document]
solidDoWant: Updated the `WithPrometheusListener` doc to explicitly state "The caller is responsible for ensuring the listener address matches `METRICS_ADDR`." Since this option is test-only and tests already set `METRICS_ADDR` to `l.Addr().String()`, a runtime equality check would add noise without safety benefit (and would break on valid `host:0` vs resolved address differences).

pkg/metrics/metrics_test.go
```diff
@@ -0,0 +1,156 @@
...
+	p, err := metrics.New(t.Context(), metrics.WithPrometheusListener(l))
+	require.NoError(t, err)
+	defer p.Shutdown(context.Background()) //nolint:errcheck
+
+	// The server is already listening on l; no poll needed.
+	url := "http://" + l.Addr().String() + "/metrics"
+	client := &http.Client{Timeout: 5 * time.Second}
+	resp, err := client.Get(url) //nolint:noctx
+	require.NoError(t, err)
+	t.Cleanup(func() { require.NoError(t, resp.Body.Close()) })
```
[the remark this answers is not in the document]
solidDoWant: Fixed — `p.Shutdown` now uses a 5-second bounded context, and `resp.Body.Close` uses `defer` (not `t.Cleanup`) so it runs before `Shutdown` in LIFO order.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,140 @@
...
+func New(ctx context.Context, opts ...Option) (*Provider, error) {
+	cfg := &config{}
+	for _, o := range opts {
+		o(cfg)
+	}
+
+	var readers []sdkmetric.Reader
+	var shutdownFuncs []func(context.Context) error
+
+	metricsAddr := os.Getenv("METRICS_ADDR")
```
solidDoWant: Unless there's a strong reason to not do this, I'd prefer if this is passed in as a parameter or option. I generally prefer for env vars to be raised as close to `main` as possible.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	if len(readers) == 0 {
+		return &Provider{
+			meterProvider: noop.NewMeterProvider(),
+			shutdown:      func(context.Context) error { return nil },
+		}, nil
+	}
+
+	opts2 := make([]sdkmetric.Option, 0, len(readers))
```
solidDoWant: name this something more meaningful

pkg/metrics/metrics.go
solidDoWant: Done — introduced an exported `Config` struct with `MetricsAddr` and `OTLPEndpoint` fields. `New` now takes `cfg Config` as a plain parameter, and `cmd/worker/main.go` reads the env vars and passes them in.

pkg/metrics/metrics.go
solidDoWant: Renamed to `sdkOpts`.

cmd/worker/main.go
```diff
@@ -28,6 +30,21 @@ func main() {
 }
 
 func run(ctx context.Context) error {
+	metricsProvider, err := metrics.New(ctx, metrics.Config{
+		MetricsAddr:  os.Getenv("METRICS_ADDR"),
+		OTLPEndpoint: os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT"),
```
solidDoWant: Because these are optional, would it be better to use an "optional" constructor pattern here?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,147 @@
...
+
+	if cfg.MetricsAddr != "" {
+		reg := prometheus.NewRegistry()
+		promReader, err := prometheusexporter.New(prometheusexporter.WithRegisterer(reg))
+		if err != nil {
+			return nil, fmt.Errorf("create prometheus exporter: %w", err)
+		}
+		readers = append(readers, promReader)
+
+		l := icfg.prometheusListener
```
solidDoWant: rename this to something more meaningful

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,147 @@
...
+	icfg := &internalConfig{}
+	for _, o := range opts {
+		o(icfg)
+	}
+
+	var readers []sdkmetric.Reader
+	var shutdownFuncs []func(context.Context) error
+
+	if cfg.MetricsAddr != "" {
+		reg := prometheus.NewRegistry()
```
solidDoWant: rename this to something more meaningful

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,147 @@
...
+func WithPrometheusListener(l net.Listener) Option {
+	return func(c *internalConfig) {
+		c.prometheusListener = l
+	}
+}
+
+// New creates a Provider from cfg. If neither MetricsAddr nor OTLPEndpoint is set,
+// a no-op MeterProvider is returned.
+func New(ctx context.Context, cfg Config, opts ...Option) (*Provider, error) {
+	icfg := &internalConfig{}
```
solidDoWant: Why is this type needed?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,147 @@
...
+
+// WithPrometheusListener supplies a pre-bound listener for the Prometheus HTTP server.
+// Config.MetricsAddr must be non-empty to enable the server; the provided listener is
+// used instead of binding a new one. The caller is responsible for ensuring the listener
+// address matches Config.MetricsAddr. Primarily useful in tests to eliminate TOCTOU races.
+func WithPrometheusListener(l net.Listener) Option {
+	return func(c *internalConfig) {
+		c.prometheusListener = l
+	}
+}
```
solidDoWant: Is this still necessary?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	if cfg.OTLPEndpoint != "" {
+		otlpExporter, err := otlpmetricgrpc.New(ctx, otlpmetricgrpc.WithEndpointURL(cfg.OTLPEndpoint))
+		if err != nil {
+			// Shut down any already-started servers before returning. Use a fresh context
+			// because ctx may already be cancelled (a common reason for init failure).
+			cleanupCtx, cleanupCancel := context.WithTimeout(context.Background(), 5*time.Second)
+			defer cleanupCancel()
+			for _, fn := range shutdownFuncs {
```
solidDoWant: fn -> shutdownFunc

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,147 @@
...
+				_ = fn(cleanupCtx)
```
solidDoWant: Should this error be captured and aggregated?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,147 @@
...
+
+	if len(readers) == 0 {
+		return &Provider{
+			meterProvider: noop.NewMeterProvider(),
+			shutdown:      func(context.Context) error { return nil },
+		}, nil
+	}
+
+	sdkOpts := make([]sdkmetric.Option, 0, len(readers))
+	for _, r := range readers {
```
solidDoWant: r -> readers

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,147 @@
...
+		sdkOpts = append(sdkOpts, sdkmetric.WithReader(r))
+	}
+	mp := sdkmetric.NewMeterProvider(sdkOpts...)
```
solidDoWant: rename this to something more meaningful

cmd/worker/main.go
solidDoWant: Done — switched to all-functional-options. `main.go` now conditionally appends `WithMetricsAddr`/`WithOTLPEndpoint` only when the env vars are non-empty, and calls `metrics.New(ctx, metricsOpts...)`.

pkg/metrics/metrics.go
solidDoWant: Renamed to `listener`.

pkg/metrics/metrics.go
solidDoWant: Renamed to `promRegistry`.

pkg/metrics/metrics.go
solidDoWant: Eliminated — `internalConfig` was only needed to separate the "internal" listener option from the exported `Config` fields. Now that all configuration flows through a single `config` struct with functional options, there's no longer a reason for two types.

pkg/metrics/metrics.go
solidDoWant: Updated — the doc now says "WithMetricsAddr must also be supplied to enable the server" instead of referencing the old `Config.MetricsAddr` field.

pkg/metrics/metrics.go
solidDoWant: Renamed to `shutdownFunc`.

pkg/metrics/metrics.go
solidDoWant: Yes — cleanup errors are now collected into `cleanupErrs` and returned via `errors.Join(append(cleanupErrs, fmt.Errorf(...))...)` so no error is silently discarded.

pkg/metrics/metrics.go
solidDoWant: Renamed to `reader`.

pkg/metrics/metrics.go
solidDoWant: Renamed to `sdkProvider`.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// WithPrometheusListener supplies a pre-bound listener for the Prometheus HTTP server.
+// WithMetricsAddr must also be supplied to enable the server; the provided listener is
+// used instead of binding a new one. The caller is responsible for ensuring the listener
+// address matches the addr passed to WithMetricsAddr. Primarily useful in tests to
+// eliminate TOCTOU races.
+func WithPrometheusListener(l net.Listener) Option {
+	return func(c *config) { c.prometheusListener = l }
+}
```
solidDoWant: Does the provider still need to support a listener like this, or can it be eliminated completely (with tests updated)?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,150 @@
...
+
+// New creates a Provider. If neither WithMetricsAddr nor WithOTLPEndpoint is supplied,
+// a no-op MeterProvider is returned.
+func New(ctx context.Context, opts ...Option) (*Provider, error) {
+	cfg := &config{}
+	for _, o := range opts {
```
solidDoWant: o -> opts

pkg/metrics/metrics.go
solidDoWant: Eliminated — removed `WithPrometheusListener` and the `prometheusListener` config field entirely. Tests now use a `freeAddr()` helper that binds a listener to get a free port, closes it, and returns the address string; `New` binds normally via `net.Listen`.

pkg/metrics/metrics.go
solidDoWant: Renamed to `opt`.

PR #59, opened 2026-03-29
feat(worker): instrument workflow runs with per-run metrics

### LINKED ISSUE #55: feat(worker): instrument workflow runs with per-run metrics
## Summary

Wire the metrics infrastructure into the media workflow, recording per-run observations for all required metrics under the standard label set, with opt-in high-cardinality labels and a dedicated counter for invalid-media files.

## Technical Context

Parent issue: #50

Depends on:
- #52 — `MediaInfo` interface and `ArrLibrary.GetInfo`
- #53 — `ProbeOutput.DurationSeconds`, `TranscodeOutput`, and file size measurements
- #54 — metrics infrastructure (`MeterProvider`, Prometheus/OTLP wiring)

### Standard label set (all processing metrics)

`source_codec`, `destination_codec`, `source_container`, `destination_container`, `media_type`, `mapping_name`, `hardware_accelerated`

### High-cardinality labels (opt-in via `METRICS_HIGH_CARDINALITY_LABELS=true`)

Both media types: `id`, `title`, `year`
Episodes additionally: `series_title`, `season_number`, `episode_number`

Labels not applicable to a given `media_type` are omitted (not set to empty string).

### Metrics to record (all are histograms unless noted)

| Metric | Unit | Source |
|---|---|---|
| `media_workflow_audio_track_count` | count | `len(ProbeOutput.AudioStreams)` |
| `media_workflow_subtitle_track_count` | count | `len(ProbeOutput.SubtitleStreams)` |
| `media_workflow_source_duration_seconds` | seconds | `ProbeOutput.DurationSeconds` |
| `media_workflow_source_file_size_bytes` | bytes | source file stat |
| `media_workflow_destination_file_size_bytes` | bytes | dest file stat |
| `media_workflow_transcode_duration_seconds` | seconds | wall-clock time of `RunTranscode` |
| `media_workflow_total_duration_seconds` | seconds | wall-clock time probe→cleanup |
| `media_workflow_invalid_files_total` | counter | incremented when `IsValidMedia=false` |

`media_workflow_invalid_files_total` carries only `media_type` and `mapping_name` labels (not the full standard set, since no transcode data is available).

### Implementation notes

- `source_codec` = `ProbeOutput.VideoCodec`; `destination_codec` = `TranscodeOutput.DestCodec`; `source_container` = `ProbeOutput.Format`; `destination_container` = `TranscodeOutput.DestContainer`.
- `hardware_accelerated` = `"true"` when `cfg.HardwareDevicePath != ""`, `"false"` otherwise.
- High-cardinality label data comes from `ArrLibrary.GetInfo` (added in #52). Call it from the notify step (or a new dedicated step) and pass the result to the metrics recorder.
- `METRICS_HIGH_CARDINALITY_LABELS` defaults to `false` if unset.

## Acceptance Criteria

- [x] **Given** a completed workflow run for a valid media file, **when** metrics are observed, **then** a single observation is recorded for each of: `media_workflow_audio_track_count`, `media_workflow_subtitle_track_count`, `media_workflow_source_duration_seconds`, `media_workflow_source_file_size_bytes`, `media_workflow_destination_file_size_bytes`, `media_workflow_transcode_duration_seconds`, `media_workflow_total_duration_seconds`.
- [x] **Given** a workflow run where the probe determines the file is not valid media, **when** metrics are observed, **then** `media_workflow_invalid_files_total` is incremented with `media_type` and `mapping_name` labels, and none of the processing metrics above are emitted for that run.
- [x] **Given** `METRICS_HIGH_CARDINALITY_LABELS` is unset or `false`, **when** metrics are observed, **then** no high-cardinality labels (`id`, `title`, `year`, `series_title`, `season_number`, `episode_number`) are present on any metric.
- [x] **Given** `METRICS_HIGH_CARDINALITY_LABELS=true`, **when** metrics are observed for a completed valid-media movie run, **then** each processing metric observation includes `id`, `title`, and `year` labels.
- [x] **Given** `METRICS_HIGH_CARDINALITY_LABELS=true`, **when** metrics are observed for a completed valid-media episode run, **then** each processing metric observation includes `id`, `title`, `year`, `series_title`, `season_number`, and `episode_number` labels.

## Non-Goals

- Metrics infrastructure setup (sub-issue #54).
- Watcher metrics.
- Tracing.
- `directory-scan` workflow metrics.
- Backfilling historical run data.

solidDoWant:
Fixes #55

### DIFF
--- a/workflows/media/media.go
+++ b/workflows/media/media.go
@@ -61,11 +61,11 @@ func NewMediaWorkflow(
 	sonarrClient medialib.ArrLibrary,
 	webhookClient *webhook.Client,
 ) *hatchet.Workflow {
-	mp := cfg.MeterProvider
-	if mp == nil {
-		mp = noop.NewMeterProvider()
+	meterProvider := cfg.MeterProvider
+	if meterProvider == nil {
+		meterProvider = noop.NewMeterProvider()
 	}
-	recorder, err := NewRecorder(mp, cfg.HighCardinalityLabels)
+	recorder, err := NewRecorder(meterProvider, cfg.HighCardinalityLabels)
 	if err != nil {
 		// Instrument registration errors are non-fatal: fall back to a noop recorder
 		// so the workflow can still run without metrics.
@@ -153,6 +153,9 @@ func NewMediaWorkflow(
 			if libErr == nil {
 				info, infoErr := library.GetInfo(ctx, input.FilePath)
 				if infoErr != nil {
+					// GetInfo failure is best-effort: log it and count it, but do not
+					// return an error. The step continues and records metrics without
+					// high-cardinality labels rather than failing the workflow run.
 					recorder.RecordMetricsError(ctx, fmt.Errorf("GetInfo: %w", infoErr))
 				} else {
 					mediaInfo = info
--- a/workflows/media/media.go
+++ b/workflows/media/media.go
@@ -4,6 +4,7 @@ package media
 
 import (
 	"fmt"
+	"log"
 	"time"
 
 	"github.com/hatchet-dev/hatchet/pkg/client/types"
@@ -67,10 +68,17 @@ func NewMediaWorkflow(
 	}
 	recorder, err := NewRecorder(meterProvider, cfg.HighCardinalityLabels)
 	if err != nil {
-		// Instrument registration errors are non-fatal: fall back to a noop recorder
-		// so the workflow can still run without metrics.
-		noopRec, _ := NewRecorder(noop.NewMeterProvider(), false)
-		recorder = noopRec
+		// Instrument registration errors are non-fatal: log for observability and fall
+		// back to a noop recorder so the workflow can still run without metrics.
+		log.Printf("media: failed to create metrics recorder: %v; falling back to noop", err)
+		var noopErr error
+		recorder, noopErr = NewRecorder(noop.NewMeterProvider(), false)
+		if noopErr != nil {
+			// noop.NewMeterProvider() instrument registration never returns errors in
+			// practice. If it somehow does, there is no safe fallback — panic to surface
+			// the bug immediately rather than silently proceeding with a nil recorder.
+			panic(fmt.Sprintf("media: failed to create noop metrics recorder: %v", noopErr))
+		}
 	}
 
 	maxRuns := int32(1)
--- a/workflows/media/media.go
+++ b/workflows/media/media.go
@@ -4,9 +4,12 @@ package media
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/hatchet-dev/hatchet/pkg/client/types"
 	hatchet "github.com/hatchet-dev/hatchet/sdks/go"
+	otelmetric "go.opentelemetry.io/otel/metric"
+	"go.opentelemetry.io/otel/metric/noop"
 
 	"github.com/solidDoWant/media-processor/pkg/medialib"
 	"github.com/solidDoWant/media-processor/pkg/webhook"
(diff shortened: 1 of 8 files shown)


workflows/media/media.go
```diff
@@ -51,6 +61,18 @@ func NewMediaWorkflow(
 	sonarrClient medialib.ArrLibrary,
 	webhookClient *webhook.Client,
 ) *hatchet.Workflow {
+	mp := cfg.MeterProvider
```
solidDoWant: name this something more meaningful

workflows/media/media.go
```diff
@@ -102,10 +131,49 @@ func NewMediaWorkflow(
...
+			return struct{}{}, fmt.Errorf("get transcode output for metrics: %w", err)
+		}
+
+		var mediaInfo medialib.MediaInfo
+		if cfg.HighCardinalityLabels {
+			library, libErr := getArrLibrary(input.MediaType, radarrClient, sonarrClient)
+			if libErr == nil {
+				info, infoErr := library.GetInfo(ctx, input.FilePath)
+				if infoErr != nil {
+					recorder.RecordMetricsError(ctx, fmt.Errorf("GetInfo: %w", infoErr))
```
solidDoWant, line 167: should this return here?

workflows/media/media.go
solidDoWant: Renamed to `meterProvider`. Fixed in fca67b8.

workflows/media/media.go
solidDoWant: Intentionally not returning here — metrics recording is best-effort. A `GetInfo` failure should not fail the `record_metrics` step or the overall workflow run; we log it, increment the `media_workflow_metrics_errors_total` counter, and continue to emit the standard-label observations without the high-cardinality labels. Added a comment in fca67b8 to make this explicit.

workflows/media/media.go
```diff
@@ -102,10 +131,52 @@ func NewMediaWorkflow(
...
+					// high-cardinality labels rather than failing the workflow run.
+					recorder.RecordMetricsError(ctx, fmt.Errorf("GetInfo: %w", infoErr))
+				} else {
+					mediaInfo = info
+				}
+			}
+		}
+
+		hardwareAccelerated := cfg.HardwareDevicePath != ""
+		totalElapsed := time.Since(probe.StartedAt)
```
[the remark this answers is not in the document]
solidDoWant: Does Hatchet provide an API for measuring task duration?

workflows/media/media.go
```diff
@@ -51,6 +61,18 @@ func NewMediaWorkflow(
...
+	meterProvider := cfg.MeterProvider
+	if meterProvider == nil {
+		meterProvider = noop.NewMeterProvider()
+	}
+	recorder, err := NewRecorder(meterProvider, cfg.HighCardinalityLabels)
+	if err != nil {
+		// Instrument registration errors are non-fatal: fall back to a noop recorder
+		// so the workflow can still run without metrics.
+		noopRec, _ := NewRecorder(noop.NewMeterProvider(), false)
+		recorder = noopRec
```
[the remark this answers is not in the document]
solidDoWant: Fixed in 3dc4592. The original error is now logged via `log.Printf` before the fallback. If the noop recorder creation itself fails (which can't happen in practice — `noop.NewMeterProvider()` never returns errors from instrument registration), we now panic immediately with a diagnostic message rather than silently proceeding with a nil recorder.

workflows/media/media.go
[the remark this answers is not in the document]
solidDoWant: No — Hatchet's `Context` interface does not expose step or workflow run timestamps. The available API is: `StepRunId()`, `WorkflowRunId()`, and `ParentOutput()` for reading step outputs. There are no start-time or elapsed-time accessors on the context. So `time.Since(probe.StartedAt)` using the serialized `time.Time` is the best available option. The Copilot concern about monotonic clock loss is valid but acceptable for an observability metric — wall-clock accuracy is sufficient for a histogram that tracks order-of-magnitude differences in run duration.

PR #61, opened 2026-03-29
feat(watcher): expose scan and dispatch metrics

### LINKED ISSUE #60: feat(watcher): expose scan and dispatch metrics
## Summary

The watcher process currently exposes no metrics. This issue adds OTel-based counters,
histograms, and a gauge covering scan execution and per-mapping dispatch activity, using
the same `pkg/metrics` infrastructure and environment variables as the worker.

## Acceptance Criteria

- [x] **Given** `METRICS_ADDR` is set and the watcher is running, **when** `/metrics` is
  scraped after at least one scan has completed, **then** `watcher_scans_total`,
  `watcher_scan_duration_seconds`, and `watcher_last_successful_scan_unix_seconds` are
  present in the output.
- [x] **Given** a mapping's walk completes without errors, **when** `/metrics` is scraped,
  **then** `watcher_scans_total{status="success", mapping_name="..."}` has incremented by 1.
- [x] **Given** a mapping's walk completes with at least one error (walk or dispatch),
  **when** `/metrics` is scraped, **then** `watcher_scans_total{status="error",
  mapping_name="..."}` has incremented by 1.
- [x] **Given** a scan completes, **when** `/metrics` is scraped, **then**
  `watcher_scan_duration_seconds` contains one observation per mapping covering that
  mapping's wall-clock walk duration, labelled with `mapping_name`.
- [x] **Given** a mapping's walk completes without errors, **when** `/metrics` is scraped,
  **then** `watcher_last_successful_scan_unix_seconds{mapping_name="..."}` is set to a
  non-zero Unix timestamp.
- [x] **Given** files are found during a scan, **when** `/metrics` is scraped, **then**
  `watcher_files_discovered_total` has incremented by the number of files found, with
  correct `mapping_name` and `media_type` labels per mapping.
- [x] **Given** a file is successfully dispatched to Hatchet, **when** `/metrics` is scraped,
  **then** `watcher_dispatches_total` has incremented by 1 with the correct `mapping_name`
  and `media_type` labels.
- [x] **Given** a dispatch call to Hatchet fails, **when** `/metrics` is scraped, **then**
  `watcher_dispatch_errors_total` has incremented by 1 with the correct `mapping_name` and
  `media_type` labels.
- [x] **Given** neither `METRICS_ADDR` nor `OTEL_EXPORTER_OTLP_ENDPOINT` is set, **when**
  the watcher starts, **then** it starts successfully and all scan/dispatch operations
  complete normally.

## Technical Context

- Depends on: #58 (metrics infrastructure PR that introduced `pkg/metrics`)
- Reuse `pkg/metrics.New` with the same env vars as the worker: `METRICS_ADDR` (Prometheus
  pull) and `OTEL_EXPORTER_OTLP_ENDPOINT` (OTLP push). Initialize in `run()` in
  `cmd/watcher/main.go` and pass the `MeterProvider` into `NewScanWorkflow`.
- Unlike the media workflow recorder (which is created per workflow run), instruments must
  be registered **once at startup** and reused across scan invocations — this is the standard
  OTel pattern for a long-lived process.
- `scan()` in `watcher.go` loops over mappings individually; per-mapping timing and status
  require tracking errors per loop iteration rather than in one shared slice, and wrapping
  each `filepath.WalkDir` call with a timer.
- Metric names and labels to implement:
  - `watcher_scans_total` — counter, labels: `mapping_name`, `status` (`success` or `error`)
  - `watcher_scan_duration_seconds` — histogram, label: `mapping_name`
  - `watcher_last_successful_scan_unix_seconds` — gauge (set, not incremented), label: `mapping_name`
  - `watcher_files_discovered_total` — counter, labels: `mapping_name`, `media_type`
  - `watcher_dispatches_total` — counter, labels: `mapping_name`, `media_type`
  - `watcher_dispatch_errors_total` — counter, labels: `mapping_name`, `media_type`
- `mapping_name` matches the label already used in worker metrics, enabling cross-process
  correlation.
- `Provider.Shutdown` must be called (deferred) in `run()` so buffered OTLP metrics are
  flushed on graceful shutdown.

## Non-Goals

- Metrics for watcher startup/config-validation failures.
- Tracking whether Hatchet deduplicated a dispatch via the idempotency key — that
  information is not observable from the watcher side.
- Any changes to the worker or `pkg/metrics`.

solidDoWant:
Instruments the watcher process with OTel metrics using the same `pkg/metrics` infrastructure and environment variables (`METRICS_ADDR`, `OTEL_EXPORTER_OTLP_ENDPOINT`) as the worker.

## Changes

- **`cmd/watcher/main.go`**: Initializes a `pkg/metrics.Provider` at startup, defers `Shutdown`, passes `MeterProvider` to `NewScanWorkflow`.
- **`cmd/watcher/watcher.go`**: Adds `scanInstruments` struct with 6 instruments registered once at startup; updates `NewScanWorkflow` to accept a `MeterProvider`; updates `scan()` to record per-mapping timing, status, file counts, and dispatch outcomes.
- **`cmd/watcher/watcher_test.go`**: Updates existing `scan()` calls to pass noop instruments; adds 8 new acceptance tests backed by `sdkmetric.ManualReader`.

Fixes #60

### DIFF
--- a/cmd/watcher/main.go
+++ b/cmd/watcher/main.go
@@ -44,14 +44,7 @@ func run(ctx context.Context, configPath string) error {
 		return fmt.Errorf("HATCHET_CLIENT_TOKEN is not set")
 	}
 
-	var metricsOpts []metrics.Option
-	if addr := os.Getenv("METRICS_ADDR"); addr != "" {
-		metricsOpts = append(metricsOpts, metrics.WithMetricsAddr(addr))
-	}
-	if endpoint := os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT"); endpoint != "" {
-		metricsOpts = append(metricsOpts, metrics.WithOTLPEndpoint(endpoint))
-	}
-	metricsProvider, err := metrics.New(ctx, metricsOpts...)
+	metricsProvider, err := metrics.NewFromEnv(ctx)
 	if err != nil {
 		return fmt.Errorf("init metrics: %w", err)
 	}
--- a/cmd/watcher/main.go
+++ b/cmd/watcher/main.go
@@ -8,7 +8,6 @@ import (
 	"os"
 	"os/signal"
 	"syscall"
-	"time"
 
 	hatchet "github.com/hatchet-dev/hatchet/sdks/go"
 
@@ -44,17 +43,11 @@ func run(ctx context.Context, configPath string) error {
 		return fmt.Errorf("HATCHET_CLIENT_TOKEN is not set")
 	}
 
-	metricsProvider, err := metrics.NewFromEnv(ctx)
+	metricsProvider, shutdown, err := metrics.NewFromEnv(ctx)
 	if err != nil {
 		return fmt.Errorf("init metrics: %w", err)
 	}
-	defer func() {
-		shutdownCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
-		defer cancel()
-		if err := metricsProvider.Shutdown(shutdownCtx); err != nil {
-			log.Printf("metrics shutdown error: %v", err)
-		}
-	}()
+	defer shutdown()
 
 	client, err := hatchet.NewClient()
 	if err != nil {
--- a/cmd/watcher/main.go
+++ b/cmd/watcher/main.go
@@ -8,8 +8,11 @@ import (
 	"os"
 	"os/signal"
 	"syscall"
+	"time"
 
 	hatchet "github.com/hatchet-dev/hatchet/sdks/go"
+
+	"github.com/solidDoWant/media-processor/pkg/metrics"
 )
 
 func main() {
@@ -41,6 +44,25 @@ func run(ctx context.Context, configPath string) error {
 		return fmt.Errorf("HATCHET_CLIENT_TOKEN is not set")
 	}
 
+	var metricsOpts []metrics.Option
+	if addr := os.Getenv("METRICS_ADDR"); addr != "" {
+		metricsOpts = append(metricsOpts, metrics.WithMetricsAddr(addr))
+	}
+	if endpoint := os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT"); endpoint != "" {
+		metricsOpts = append(metricsOpts, metrics.WithOTLPEndpoint(endpoint))
+	}
+	metricsProvider, err := metrics.New(ctx, metricsOpts...)
+	if err != nil {
+		return fmt.Errorf("init metrics: %w", err)
+	}
+	defer func() {
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+		defer cancel()
+		if err := metricsProvider.Shutdown(shutdownCtx); err != nil {
+			log.Printf("metrics shutdown error: %v", err)
+		}
+	}()
+
 	client, err := hatchet.NewClient()
 	if err != nil {
 		return fmt.Errorf("connect to Hatchet: %w", err)
@@ -48,7 +70,10 @@ func run(ctx context.Context, configPath string) error {
 
 	log.Println("connected to Hatchet")
 
-	scanWorkflow := NewScanWorkflow(client, cfg)
+	scanWorkflow, err := NewScanWorkflow(client, cfg, metricsProvider.MeterProvider())
+	if err != nil {
+		return fmt.Errorf("create scan workflow: %w", err)
+	}
 
 	worker, err := client.NewWorker("mediaprocessor-watcher",
 		hatchet.WithWorkflows(scanWorkflow),
--- a/cmd/watcher/watcher.go
+++ b/cmd/watcher/watcher.go
@@ -7,16 +7,94 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"time"
 
 	"github.com/hatchet-dev/hatchet/pkg/client/types"
 	hatchet "github.com/hatchet-dev/hatchet/sdks/go"
+	"go.opentelemetry.io/otel/attribute"
+	otelmetric "go.opentelemetry.io/otel/metric"
 
 	"github.com/solidDoWant/media-processor/pkg/medialib"
 	"github.com/solidDoWant/media-processor/workflows/media"
 )
 
+func mappingNameAttr(name string) attribute.KeyValue {
+	return attribute.String("mapping_name", name)
+}
+
+func mediaTypeAttr(mt medialib.MediaType) attribute.KeyValue {
+	return attribute.String("media_type", string(mt))
+}
+
+func statusAttr(status string) attribute.KeyValue {
+	return attribute.String("status", status)
+}
+
 // dispatchFunc submits a workflow run for the given absolute file path, media type, and mapping name.
 type dispatchFunc func(ctx context.Context, filePath string, mediaType medialib.MediaType, mappingName string) error
+
+// scanInstruments holds all OTel instruments used during scan. Instruments are registered
+// once at startup and reused across every scan invocation.
+type scanInstruments struct {
+	scansTotal           otelmetric.Int64Counter
+	scanDuration         otelmetric.Float64Histogram
+	lastSuccessfulScan   otelmetric.Float64Gauge
+	filesDiscoveredTotal otelmetric.Int64Counter
+	dispatchesTotal      otelmetric.Int64Counter
+	dispatchErrorsTotal  otelmetric.Int64Counter
+}
+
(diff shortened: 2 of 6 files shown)


cmd/watcher/main.go
```diff
@@ -41,14 +44,36 @@ func run(ctx context.Context, configPath string) error {
...
+		return fmt.Errorf("init metrics: %w", err)
+	}
+	defer func() {
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+		defer cancel()
+		if err := metricsProvider.Shutdown(shutdownCtx); err != nil {
+			log.Printf("metrics shutdown error: %v", err)
+		}
+	}()
+
```
solidDoWant: Both this and the the worker use very similar logic here. Do you have an opinion on pulling this out into a separate function in the metrics package?

cmd/watcher/watcher.go
```diff
@@ -7,28 +7,111 @@ import (
...
+	scanDuration         otelmetric.Float64Histogram
+	lastSuccessfulScan   otelmetric.Float64Gauge
+	filesDiscoveredTotal otelmetric.Int64Counter
+	dispatchesTotal      otelmetric.Int64Counter
+	dispatchErrorsTotal  otelmetric.Int64Counter
+}
+
+// newScanInstruments registers all watcher scan instruments with the given MeterProvider.
+func newScanInstruments(mp otelmetric.MeterProvider) (*scanInstruments, error) {
+	meter := mp.Meter("github.com/solidDoWant/media-processor/cmd/watcher")
```
solidDoWant: Is there any feasible way to get this without hard-coding it? This seems kinda brittle.

cmd/watcher/watcher.go
```diff
@@ -44,7 +127,7 @@ func NewScanWorkflow(client *hatchet.Client, cfg *Config) *hatchet.StandaloneTas
 				)
 				return err
 			}
-			return struct{}{}, scan(ctx, cfg, dispatch)
+			return struct{}{}, scan(ctx, cfg, instruments, dispatch)
```
solidDoWant, line 139: The media workflow uses a separate task for this. Do you have an opinion on this approach?

cmd/watcher/main.go
solidDoWant: Agreed — both `run()` functions share the same three-line env-var → opts → `metrics.New` pattern, and a `metrics.NewFromEnv(ctx)` helper in `pkg/metrics` would eliminate the duplication. This PR's Non-Goals explicitly exclude changes to `pkg/metrics`, so I've left it as-is here. I'll file a follow-up issue to add that helper.

cmd/watcher/watcher.go
solidDoWant: In Go there's no compile-time way to retrieve the current package's import path, so hardcoding is the idiomatic OTel convention here. To make it less brittle I've extracted the string to a named constant (`meterName`) in the same file — it's now easy to find, rename, or grep for. Just pushed.

cmd/watcher/watcher.go
solidDoWant: My opinion is that the inline approach is the right fit here. The media workflow's `Recorder` type earns its existence because recording spans multiple workflow tasks that each call different record methods, and it needs to be passed between them. In the watcher there's a single recording site (`scan()`), so introducing a separate type would add indirection without a meaningful benefit. Happy to extract it if you feel strongly about consistency, but I'd lean toward keeping it simple.

cmd/watcher/main.go
solidDoWant: It's okay, you can add this specifically.

solidDoWant:
Where are metric pushes handled?

cmd/watcher/main.go
solidDoWant: Done — added `NewFromEnv(ctx)` to `pkg/metrics` and updated both `cmd/watcher/main.go` and `cmd/worker/main.go` to use it. Also added three tests covering the no-env, METRICS_ADDR, and OTLP_ENDPOINT cases.

solidDoWant:
Metric pushes are handled automatically by the OTel SDK. When `OTEL_EXPORTER_OTLP_ENDPOINT` is set, `pkg/metrics.New` (now called via `NewFromEnv`) creates a `sdkmetric.PeriodicReader` wrapping an OTLP gRPC exporter. The SDK's periodic reader runs a background goroutine that flushes buffered metrics to the collector on a fixed interval (default 60 s). No explicit push calls are needed in application code. On shutdown, `Provider.Shutdown` triggers a final flush before the process exits.

cmd/watcher/main.go
```diff
@@ -41,14 +44,29 @@ func run(ctx context.Context, configPath string) error {
...
```
solidDoWant: I think the shutdown closure could be moved to the `NewFromEnv` constructor, and returned.

solidDoWant: Done. `NewFromEnv` now returns `(*Provider, func(), error)` — the shutdown func encapsulates the 10-second timeout context and stderr error logging, so callers reduce to `defer shutdown()`. Just pushed.

cmd/watcher/watcher.go
```diff
@@ -106,12 +200,20 @@ func scan(ctx context.Context, cfg *Config, dispatch dispatchFunc) error {
...
+				otelmetric.WithAttributes(mappingNameAttr(w.Name), mediaTypeAttr(w.MediaType)))
+
 			if dispatchErr := dispatch(ctx, absPath, w.MediaType, w.Name); dispatchErr != nil {
-				errs = append(errs, fmt.Errorf("dispatch workflow for %q (media type %v): %w", absPath, w.MediaType, dispatchErr))
+				mappingErrs = append(mappingErrs, fmt.Errorf("dispatch workflow for %q (media type %v): %w", absPath, w.MediaType, dispatchErr))
+				instruments.dispatchErrorsTotal.Add(ctx, 1,
+					otelmetric.WithAttributes(mappingNameAttr(w.Name), mediaTypeAttr(w.MediaType)))
+			} else {
+				instruments.dispatchesTotal.Add(ctx, 1,
+					otelmetric.WithAttributes(mappingNameAttr(w.Name), mediaTypeAttr(w.MediaType)))
```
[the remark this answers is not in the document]
solidDoWant: Good catch. Precomputed `mappingOpt`, `fileOpt`, `successOpt`, and `errorOpt` once per watch entry before the `WalkDir` call — the callback now reuses the same option objects rather than re-allocating on every file. Just pushed.

cmd/watcher/watcher.go
```diff
@@ -120,7 +222,20 @@ func scan(ctx context.Context, cfg *Config, dispatch dispatchFunc) error {
...
+		duration := time.Since(start).Seconds()
+		instruments.scanDuration.Record(ctx, duration, mappingAttr)
+
+		if len(mappingErrs) == 0 {
+			instruments.scansTotal.Add(ctx, 1,
+				otelmetric.WithAttributes(mappingNameAttr(w.Name), statusAttr("success")))
+			instruments.lastSuccessfulScan.Record(ctx, float64(time.Now().Unix()), mappingAttr)
+		} else {
+			instruments.scansTotal.Add(ctx, 1,
+				otelmetric.WithAttributes(mappingNameAttr(w.Name), statusAttr("error")))
```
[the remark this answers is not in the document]
solidDoWant: Done — added `scanStatusSuccess` and `scanStatusError` constants and replaced the inline string literals. Just pushed.

pkg/metrics/metrics_test.go
```diff
@@ -129,6 +129,43 @@ func TestBothExporters_Active(t *testing.T) {
...
+func TestNewFromEnv_OTLPEndpoint_CreatesProvider(t *testing.T) {
+	l := bindListener(t)
+	t.Setenv("METRICS_ADDR", "")
+	t.Setenv("OTEL_EXPORTER_OTLP_ENDPOINT", "http://"+l.Addr().String())
+
+	p, shutdown, err := metrics.NewFromEnv(t.Context())
+	require.NoError(t, err)
+	t.Cleanup(shutdown)
+	require.NotNil(t, p.MeterProvider())
+}
```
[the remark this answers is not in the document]
solidDoWant: Good point. Updated `TestNewFromEnv_OTLPEndpoint_CreatesProvider` to use `p.Shutdown` directly with a 1-second context deadline rather than `t.Cleanup(shutdown)`, so the test doesn't hang for up to 10 seconds waiting for a flush against the dummy listener. The `shutdown` func (with the full 10s budget) is still appropriate in production where a real collector is present. Just pushed.

PR #70, opened 2026-03-30
feat(medialib): fetch poster artwork from arr libraries and embed in output MKV

### LINKED ISSUE #69: feat(medialib): fetch poster artwork from arr libraries and embed in output MKV
## Summary

When transcoding or transmuxing media, the output MKV should have the item's poster artwork embedded as an attachment stream, sourced from the Radarr or Sonarr library that owns the file. This makes artwork available to media players (e.g. Kodi, Infuse, VLC) without requiring a separate sidecar file.

## Acceptance Criteria

- [x] **Given** a movie file is transcoded and Radarr returns a poster image for it, **when** the transcode completes, **then** the output MKV contains exactly one attachment stream with the poster image bytes and MIME type set accordingly (`image/jpeg` or `image/png`).

- [x] **Given** a TV episode file is transcoded and Sonarr returns a poster image for the series, **when** the transcode completes, **then** the output MKV contains exactly one attachment stream with the series poster image bytes and MIME type set accordingly.

- [x] **Given** the source file already contains one or more attachment streams, **when** the transcode completes, **then** those existing attachment streams are not present in the output — only the freshly fetched poster is embedded.

- [x] **Given** the arr library is unreachable or returns no poster image for the item, **when** the transcode step runs, **then** transcoding proceeds and completes successfully without an embedded attachment, a warning is logged, and the `media_workflow_artwork_fetch_skipped_total` counter is incremented.

- [x] **Given** the `ArrLibrary` interface, **when** a poster image is available for the item, **then** it is retrievable via a new `GetPosterImage` method on the interface that returns the raw image bytes and MIME type string.

## Technical Context

- Depends on: none

- **Artwork source:** `golift.io/starr` (already a dependency) returns `[]*starr.Image` on movie and series responses. Each `starr.Image` has a `CoverType` field — use `"poster"` only. The `URL` field is a relative path (e.g. `/MediaCover/123/poster.jpg`) served by the arr instance; prepend the configured base URL and include the API key header to fetch it. `RemoteURL` is a direct external URL and may be used as a fallback if `URL` is empty.

- **MKV embedding:** In the MKV container (the only output format this project produces), artwork lives in the `extradata` field of a new stream's codec parameters — no packets are written for it. The stream must be created before `WriteHeader`. Required setup via `go-astiav`:
  1. `outputFormatCtx.NewStream(nil)`
  2. `stream.CodecParameters().SetMediaType(astiav.MediaTypeAttachment)`
  3. `stream.CodecParameters().SetCodecID(astiav.CodecIDPng)` or `CodecIDMjpeg`
  4. `stream.CodecParameters().SetExtraData(imageBytes)` — uses `av_mallocz` internally, correct allocator
  5. Stream metadata dict with keys `"mimetype"` (`"image/png"` or `"image/jpeg"`) and `"filename"` (`"cover.png"` / `"cover.jpg"`)

- **Suppressing existing attachments:** The transcode pipeline currently copies existing attachment streams via `copyStreamState`. When poster bytes are available, existing attachment streams from the source file must be excluded from the output (so only the new poster is embedded).

- **Builder API extension:** Add a `WithCoverArt(imageBytes []byte, mimeType string)` method to `TranscodeBuilder` in `pkg/ffmpeg`. `RunTranscode` in `workflows/shared/transcode.go` is responsible for fetching the image and passing it in.

- **New metric:** `media_workflow_artwork_fetch_skipped_total` — an `Int64Counter` registered in `workflows/media/recorder.go`, incremented once per run where artwork fetch is attempted but yields no image (library unreachable, no poster available, or download error).

- **Where to fetch:** artwork should be fetched inside `RunTranscode` (or a helper it calls) in `workflows/shared/transcode.go`, using an `ArrLibrary` reference passed in via a new parameter or extended input struct.

## Non-Goals

- Embedding any artwork type other than `poster` (fanart, banner, clearlogo, etc.).
- Embedding artwork in MP4 output.
- Fetching artwork from any source other than the configured Radarr/Sonarr instance.
- Sourcing images from the local filesystem or any sidecar file.
- Updating existing files already in the output directory that were processed before this feature was added.

solidDoWant:
## Summary

- Add `GetPosterImage` to the `ArrLibrary` interface, implemented for Radarr (movie poster) and Sonarr (series poster) via a shared `pkg/medialib/internal/artwork` package
- Artwork is validated twice: extension pre-check (`.jpg`/`.jpeg`/`.png` only before fetching) and `Content-Type` post-fetch validation
- Add `WithCoverArt` to `TranscodeBuilder`; strips existing attached-picture streams from the input (written as `MediaTypeAttachment`, read back by the matroska demuxer with `DispositionFlagAttachedPic`) and embeds the new poster as an MKV attachment stream
- `RunTranscode` accepts an optional `ArrLibrary` and falls back gracefully with a warning + `ArtworkFetchSkipped` flag when artwork is unavailable or cannot be fetched
- Increment `media_workflow_artwork_fetch_skipped_total` OTel counter when the fallback path is taken

## Test plan

- [x] `pkg/medialib/internal/artwork`: JPEG/PNG fetch, non-poster skip, unsupported extension, RemoteURL fallback, unsupported Content-Type, unreachable server
- [x] `pkg/medialib/radarr`: JPEG/PNG poster, no poster (fanart only), unsupported extension, episode not found, unreachable URL
- [x] `pkg/medialib/sonarr`: same coverage as radarr
- [x] `pkg/ffmpeg`: JPEG/PNG embedded correctly, existing attachments stripped on re-transcode, nil bytes is no-op
- [x] `workflows/shared`: existing RunTranscode tests pass with `nil` library (no regression)
- [x] All tests pass (`go test ./...`) and lint is clean (`make lint`)

Fixes #69

### DIFF
--- a/pkg/ffmpeg/transcode.go
+++ b/pkg/ffmpeg/transcode.go
@@ -178,8 +178,9 @@ func (b *TranscodeBuilder) WithDownmixTitle(title string) *TranscodeBuilder {
 }
 
 // WithCoverArt embeds imageBytes as a cover art attachment stream in the MKV
-// output. mimeType must be "image/jpeg" or "image/png"; any other value is
-// treated as a no-op. When called with a valid MIME type, any existing
+// output. This method is a no-op when the output container is not
+// ContainerMKV. mimeType must be "image/jpeg" or "image/png"; any other value
+// is treated as a no-op. When called with a valid MIME type, any existing
 // attachment streams present in the source file are excluded from the output
 // so that only the supplied artwork is embedded. A nil or empty imageBytes
 // slice is a no-op.
@@ -392,11 +393,12 @@ func (t *Transcoder) buildStreamStates(inputFmt *astiav.FormatContext, hwAccel H
 
 		mediaType := inStream.CodecParameters().MediaType()
 
-		// When cover art is being embedded, drop all existing attachment streams
-		// so the output contains only the freshly fetched poster. MKV attachments
-		// are written as MediaTypeAttachment but read back by the matroska demuxer
-		// as video streams with DispositionFlagAttachedPic, so we must check both.
-		if len(t.coverArtBytes) > 0 {
+		// When cover art is being embedded into a Matroska output, drop all
+		// existing attachment streams so the output contains only the freshly
+		// fetched poster. MKV attachments are written as MediaTypeAttachment but
+		// read back by the matroska demuxer as video streams with
+		// DispositionFlagAttachedPic, so we must check both.
+		if len(t.coverArtBytes) > 0 && t.container == ContainerMKV {
 			if mediaType == astiav.MediaTypeAttachment {
 				continue
 			}
@@ -610,8 +612,8 @@ func (t *Transcoder) setupOutputContext(streams map[int]stream, downmix *audioSt
 		downmixOut.SetMetadata(downmixMeta)
 	}
 
-	// Add cover art attachment stream after all regular streams.
-	if len(t.coverArtBytes) > 0 {
+	// Add cover art attachment stream after all regular streams (MKV only).
+	if len(t.coverArtBytes) > 0 && t.container == ContainerMKV {
 		if err := t.addCoverArtStream(outputFmt); err != nil {
 			outputFmt.Free()
 			return nil, noopClose, err
--- a/pkg/ffmpeg/transcode.go
+++ b/pkg/ffmpeg/transcode.go
@@ -178,12 +178,19 @@ func (b *TranscodeBuilder) WithDownmixTitle(title string) *TranscodeBuilder {
 }
 
 // WithCoverArt embeds imageBytes as a cover art attachment stream in the MKV
-// output. mimeType must be "image/jpeg" or "image/png". When called, any
-// existing attachment streams present in the source file are excluded from the
-// output so that only the supplied artwork is embedded. A nil or empty
-// imageBytes slice is a no-op.
+// output. mimeType must be "image/jpeg" or "image/png"; any other value is
+// treated as a no-op. When called with a valid MIME type, any existing
+// attachment streams present in the source file are excluded from the output
+// so that only the supplied artwork is embedded. A nil or empty imageBytes
+// slice is a no-op.
 func (b *TranscodeBuilder) WithCoverArt(imageBytes []byte, mimeType string) *TranscodeBuilder {
 	if len(imageBytes) == 0 {
+		return b
+	}
+
+	switch mimeType {
+	case "image/jpeg", "image/png":
+	default:
 		return b
 	}
 
--- a/pkg/ffmpeg/transcode.go
+++ b/pkg/ffmpeg/transcode.go
@@ -211,6 +211,15 @@ func (b *TranscodeBuilder) WithDownmix(idx *int) *TranscodeBuilder {
 	return b
 }
 
+// effectiveContainerIsMKV reports whether the output container is Matroska,
+// either because it was explicitly set via ToContainer or because the output
+// filename has a .mkv extension (the case where the container is inferred by
+// astiav.AllocOutputFormatContext from the filename).
+func (b *TranscodeBuilder) effectiveContainerIsMKV() bool {
+	return b.container == ContainerMKV ||
+		(b.container == "" && strings.HasSuffix(strings.ToLower(b.outputPath), ".mkv"))
+}
+
 // Build returns a runnable Transcoder.
 func (b *TranscodeBuilder) Build() *Transcoder {
 	return &Transcoder{TranscodeBuilder: *b}
@@ -398,7 +407,7 @@ func (t *Transcoder) buildStreamStates(inputFmt *astiav.FormatContext, hwAccel H
 		// fetched poster. MKV attachments are written as MediaTypeAttachment but
 		// read back by the matroska demuxer as video streams with
 		// DispositionFlagAttachedPic, so we must check both.
-		if len(t.coverArtBytes) > 0 && t.container == ContainerMKV {
+		if len(t.coverArtBytes) > 0 && t.effectiveContainerIsMKV() {
 			if mediaType == astiav.MediaTypeAttachment {
 				continue
 			}
@@ -613,7 +622,7 @@ func (t *Transcoder) setupOutputContext(streams map[int]stream, downmix *audioSt
 	}
 
 	// Add cover art attachment stream after all regular streams (MKV only).
-	if len(t.coverArtBytes) > 0 && t.container == ContainerMKV {
+	if len(t.coverArtBytes) > 0 && t.effectiveContainerIsMKV() {
 		if err := t.addCoverArtStream(outputFmt); err != nil {
 			outputFmt.Free()
 			return nil, noopClose, err
--- a/pkg/ffmpeg/transcode.go
+++ b/pkg/ffmpeg/transcode.go
@@ -685,21 +685,26 @@ func (t *Transcoder) addCoverArtStream(outputFmt *astiav.FormatContext) error {
 	cp := artStream.CodecParameters()
 	cp.SetMediaType(astiav.MediaTypeAttachment)
 
+	var (
+		codecID  astiav.CodecID
+		filename string
+	)
+
 	switch t.coverArtMimeType {
+	case "image/jpeg":
+		codecID = astiav.CodecIDMjpeg
+		filename = "cover.jpg"
 	case "image/png":
-		cp.SetCodecID(astiav.CodecIDPng)
+		codecID = astiav.CodecIDPng
+		filename = "cover.png"
 	default:
-		// Treat image/jpeg and any unknown type as MJPEG.
-		cp.SetCodecID(astiav.CodecIDMjpeg)
+		return fmt.Errorf("ffmpeg: unsupported cover art MIME type %q", t.coverArtMimeType)
 	}
+
+	cp.SetCodecID(codecID)
 
 	if err := cp.SetExtraData(t.coverArtBytes); err != nil {
 		return fmt.Errorf("ffmpeg: setting cover art extradata: %w", err)
-	}
-
-	filename := "cover.jpg"
-	if t.coverArtMimeType == "image/png" {
-		filename = "cover.png"
 	}
 
 	meta := astiav.NewDictionary()
--- a/pkg/ffmpeg/transcode.go
+++ b/pkg/ffmpeg/transcode.go
@@ -27,6 +27,8 @@ type TranscodeBuilder struct {
 	subtitleStreamTitles  map[int]string // per-stream title overrides for subtitle streams; nil = no overrides
 	autoDownmixTitle      bool           // derive downmix stream title from actual encoder channel layout
 	downmixTitle          string         // title prefix prepended to the downmix channel layout label
+	coverArtBytes         []byte         // raw image bytes to embed as MKV attachment; nil = no cover art
+	coverArtMimeType      string         // MIME type of coverArtBytes ("image/jpeg" or "image/png")
 }
 
 // NewTranscode returns a builder for a transcode job from inputPath to outputPath.
@@ -171,6 +173,22 @@ func (b *TranscodeBuilder) WithDownmixTitle(title string) *TranscodeBuilder {
 	}
 
 	b.downmixTitle = title
+
+	return b
+}
+
+// WithCoverArt embeds imageBytes as a cover art attachment stream in the MKV
+// output. mimeType must be "image/jpeg" or "image/png". When called, any
+// existing attachment streams present in the source file are excluded from the
+// output so that only the supplied artwork is embedded. A nil or empty
+// imageBytes slice is a no-op.
+func (b *TranscodeBuilder) WithCoverArt(imageBytes []byte, mimeType string) *TranscodeBuilder {
+	if len(imageBytes) == 0 {
+		return b
+	}
+
+	b.coverArtBytes = imageBytes
+	b.coverArtMimeType = mimeType
 
 	return b
 }
@@ -366,6 +384,22 @@ func (t *Transcoder) buildStreamStates(inputFmt *astiav.FormatContext, hwAccel H
 		}
 
 		mediaType := inStream.CodecParameters().MediaType()
+
+		// When cover art is being embedded, drop all existing attachment streams
+		// so the output contains only the freshly fetched poster. MKV attachments
+		// are written as MediaTypeAttachment but read back by the matroska demuxer
+		// as video streams with DispositionFlagAttachedPic, so we must check both.
+		if len(t.coverArtBytes) > 0 {
+			if mediaType == astiav.MediaTypeAttachment {
+				continue
+			}
+
+			if mediaType == astiav.MediaTypeVideo &&
+				inStream.DispositionFlags().Has(astiav.DispositionFlagAttachedPic) {
+				continue
+			}
+		}
+
 		base := copyStreamState{inStream: inStream}
 
 		var s stream
@@ -569,6 +603,14 @@ func (t *Transcoder) setupOutputContext(streams map[int]stream, downmix *audioSt
 		downmixOut.SetMetadata(downmixMeta)
 	}
 
+	// Add cover art attachment stream after all regular streams.
+	if len(t.coverArtBytes) > 0 {
+		if err := t.addCoverArtStream(outputFmt); err != nil {
+			outputFmt.Free()
+			return nil, noopClose, err
+		}
+	}
+
 	// Open the IO context for file-based output formats.
 	closeIO := noopClose
 
@@ -611,6 +653,52 @@ func (t *Transcoder) outputDisposition(inStream *astiav.Stream) astiav.Dispositi
 	}
 
 	return disp.Del(astiav.DispositionFlagDefault)
+}
+
+// addCoverArtStream appends a cover art attachment stream to outputFmt using
+// t.coverArtBytes and t.coverArtMimeType. The image bytes are stored in the
+// stream's codec parameters extradata; no packets are written for this stream.
+func (t *Transcoder) addCoverArtStream(outputFmt *astiav.FormatContext) error {
+	artStream := outputFmt.NewStream(nil)
+	if artStream == nil {
+		return errors.New("ffmpeg: failed to create cover art attachment stream")
+	}
+
+	cp := artStream.CodecParameters()
+	cp.SetMediaType(astiav.MediaTypeAttachment)
+
+	switch t.coverArtMimeType {
+	case "image/png":
+		cp.SetCodecID(astiav.CodecIDPng)
+	default:
+		// Treat image/jpeg and any unknown type as MJPEG.
+		cp.SetCodecID(astiav.CodecIDMjpeg)
+	}
+
+	if err := cp.SetExtraData(t.coverArtBytes); err != nil {
+		return fmt.Errorf("ffmpeg: setting cover art extradata: %w", err)
+	}
+
+	filename := "cover.jpg"
+	if t.coverArtMimeType == "image/png" {
+		filename = "cover.png"
+	}
+
+	meta := astiav.NewDictionary()
+
+	if err := meta.Set("mimetype", t.coverArtMimeType, astiav.NewDictionaryFlags()); err != nil {
+		meta.Free()
+		return fmt.Errorf("ffmpeg: setting cover art mimetype metadata: %w", err)
+	}
+
+	if err := meta.Set("filename", filename, astiav.NewDictionaryFlags()); err != nil {
+		meta.Free()
+		return fmt.Errorf("ffmpeg: setting cover art filename metadata: %w", err)
+	}
+
+	artStream.SetMetadata(meta)
+
+	return nil
 }
 
 // readAllPackets is the main decode/encode loop.
--- a/pkg/medialib/internal/artwork/artwork.go
+++ b/pkg/medialib/internal/artwork/artwork.go
@@ -1,0 +1,86 @@
+// Package artwork provides shared utilities for fetching poster images from
+// arr-style media library services (Radarr, Sonarr).
+package artwork
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+
+	"golift.io/starr"
+)
+
+const (
+	// MimeJPEG is the MIME type for JPEG images.
+	MimeJPEG = "image/jpeg"
+	// MimePNG is the MIME type for PNG images.
+	MimePNG = "image/png"
+)
+
+// FetchPosterImage finds the poster image in images, fetches it from baseURL
+// using the provided API key, and returns the bytes and MIME type.
+// Returns nil bytes (no error) when no JPEG or PNG poster is available or
+// the image type cannot be validated.
+func FetchPosterImage(ctx context.Context, images []*starr.Image, baseURL, apiKey string) ([]byte, string, error) {
+	for _, img := range images {
+		if img.CoverType != "poster" {
+			continue
+		}
+
+		// Pre-fetch extension check: only accept JPEG and PNG.
+		ext := strings.ToLower(img.Extension)
+		if ext != ".jpg" && ext != ".jpeg" && ext != ".png" {
+			return nil, "", nil
+		}
+
+		imageURL := img.URL
+		if imageURL == "" {
+			imageURL = img.RemoteURL
+		}
+
+		if imageURL == "" {
+			return nil, "", nil
+		}
+
+		// Relative paths are served by the arr instance; prepend the base URL.
+		if strings.HasPrefix(imageURL, "/") {
+			imageURL = strings.TrimRight(baseURL, "/") + imageURL
+		}
+
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, imageURL, nil)
+		if err != nil {
+			return nil, "", fmt.Errorf("build image request: %w", err)
+		}
+
+		req.Header.Set("X-Api-Key", apiKey)
+
+		resp, err := http.DefaultClient.Do(req)
+		if err != nil {
+			return nil, "", fmt.Errorf("fetch poster image: %w", err)
+		}
+
+		defer func() { _ = resp.Body.Close() }()
+
+		// Post-fetch Content-Type validation.
+		ct := resp.Header.Get("Content-Type")
+		// Strip parameters like "; charset=utf-8".
+		if i := strings.Index(ct, ";"); i >= 0 {
+			ct = strings.TrimSpace(ct[:i])
+		}
+
+		if ct != MimeJPEG && ct != MimePNG {
+			return nil, "", nil
+		}
+
+		data, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return nil, "", fmt.Errorf("read poster image: %w", err)
+		}
+
+		return data, ct, nil
+	}
+
+	return nil, "", nil
+}
--- a/pkg/medialib/internal/artwork/artwork.go
+++ b/pkg/medialib/internal/artwork/artwork.go
@@ -6,8 +6,10 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"mime"
 	"net/http"
 	"strings"
+	"time"
 
 	"golift.io/starr"
 )
@@ -17,13 +19,21 @@ const (
 	MimeJPEG = "image/jpeg"
 	// MimePNG is the MIME type for PNG images.
 	MimePNG = "image/png"
+
+	// fetchTimeout is the per-request deadline for poster image downloads.
+	fetchTimeout = 10 * time.Second
 )
 
 // FetchPosterImage finds the poster image in images, fetches it from baseURL
 // using the provided API key, and returns the bytes and MIME type.
 // Returns nil bytes (no error) when no JPEG or PNG poster is available or
 // the image type cannot be validated.
+//
+// The API key is only sent when the resolved image URL starts with baseURL.
+// RemoteURL values (absolute external URLs) are fetched without the API key.
 func FetchPosterImage(ctx context.Context, images []*starr.Image, baseURL, apiKey string) ([]byte, string, error) {
+	normalizedBase := strings.TrimRight(baseURL, "/")
+
 	for _, img := range images {
 		if img.CoverType != "poster" {
 			continue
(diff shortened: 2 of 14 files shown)


pkg/medialib/internal/artwork/artwork.go
solidDoWant, line 1: Should this be an internal package?

pkg/medialib/internal/artwork/artwork.go
```diff
@@ -0,0 +1,86 @@
...
+		if imageURL == "" {
+			return nil, "", nil
+		}
+
+		// Relative paths are served by the arr instance; prepend the base URL.
+		if strings.HasPrefix(imageURL, "/") {
+			imageURL = strings.TrimRight(baseURL, "/") + imageURL
+		}
+
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, imageURL, nil)
```
solidDoWant: Add a shorter context on this; maybe 5-10s?

pkg/medialib/internal/artwork/artwork.go
```diff
@@ -0,0 +1,86 @@
...
+		if err != nil {
+			return nil, "", fmt.Errorf("build image request: %w", err)
+		}
+
+		req.Header.Set("X-Api-Key", apiKey)
```
solidDoWant, line 147: This is a major security flaw. Only send the API key when the request URL starts with the base URL configured as the service.

Evaluate this PR for other security issues.

pkg/medialib/internal/artwork/artwork.go
```diff
@@ -0,0 +1,86 @@
...
+		if err != nil {
+			return nil, "", fmt.Errorf("fetch poster image: %w", err)
+		}
+
+		defer func() { _ = resp.Body.Close() }()
+
+		// Post-fetch Content-Type validation.
+		ct := resp.Header.Get("Content-Type")
+		// Strip parameters like "; charset=utf-8".
+		if i := strings.Index(ct, ";"); i >= 0 {
```
solidDoWant: Would a parsing library be more appropriate here? I _think_ go has built-in support for this but I'm not 100% sure.

pkg/medialib/internal/artwork/artwork.go
solidDoWant: Yes, `internal` is intentional here. Go's `internal` visibility rule restricts access to the parent tree — `pkg/medialib/internal/artwork` is reachable by `pkg/medialib/radarr` and `pkg/medialib/sonarr` (both under `pkg/medialib/`) but not by any code outside that subtree. This keeps the shared helper from leaking into the public API surface.

pkg/medialib/internal/artwork/artwork.go
solidDoWant: Done — added `context.WithTimeout(ctx, 10*time.Second)` wrapping the per-image fetch (654f7a4).

pkg/medialib/internal/artwork/artwork.go
solidDoWant: Fixed in 654f7a4. The API key header is now only set when `strings.HasPrefix(imageURL, normalizedBase)` — so relative `URL` fields (prepended with `baseURL`) receive it, but `RemoteURL` values pointing to external hosts do not. A new test `TestFetchPosterImage_RemoteURLExternalHost` covers this: the external server rejects any request that carries the key, so the test fails if the key is incorrectly forwarded.

I also reviewed for other issues:
- **Redirects**: `http.DefaultClient` follows redirects and does not strip custom headers like `X-Api-Key` on cross-host redirects (only `Authorization` is stripped). However, since all callers use either the arr instance URL or a known `RemoteURL`, and the arr-served images are served as direct responses (no redirect chain), the risk is low in practice. If stricter control is needed a custom `CheckRedirect` could be added in a follow-up.
- No other security issues identified.

pkg/medialib/internal/artwork/artwork.go
solidDoWant: Switched to `mime.ParseMediaType` (654f7a4) — Go's standard library does have this in the `mime` package. It correctly handles parameters, quoted strings, and whitespace per RFC 2045, which is strictly better than manual splitting on `;`.

pkg/medialib/internal/artwork/artwork.go
solidDoWant: The remote URL may not be known ahead of time. Make sure that the API key is _exclusively_ sent when the request URL has a base of the provided URL.

solidDoWant: Fixed in 75b24f7. A custom `http.Client` with `CheckRedirect` is now used for all fetches. If a redirect target URL does not start with `normalizedBase`, the redirect is stopped (`http.ErrUseLastResponse`) — so the API key is never forwarded to an external host, even via a server-issued redirect. A new test `TestFetchPosterImage_CrossHostRedirectDoesNotSendAPIKey` verifies this: the external redirect target records whether it received the key, and the assertion fails if it did.

solidDoWant: The redirect can and should be followed, but the API key should only be sent to the explicitly-configured host.

solidDoWant: Fixed in b19e5d4. `CheckRedirect` now deletes the `X-Api-Key` header (rather than returning `ErrUseLastResponse`) when the redirect target URL does not start with `normalizedBase`. The redirect is followed normally; the key is simply absent from the request to the external host. `TestFetchPosterImage_CrossHostRedirectDoesNotSendAPIKey` now also asserts that the image is returned correctly (i.e. the redirect was followed and served the image).

pkg/medialib/internal/artwork/artwork.go
```diff
@@ -0,0 +1,113 @@
...
+		req, err := http.NewRequestWithContext(fetchCtx, http.MethodGet, imageURL, nil)
+		if err != nil {
+			return nil, "", fmt.Errorf("build image request: %w", err)
+		}
+
+		// Only send the API key when fetching from the configured arr instance.
+		// RemoteURL values may point to external CDNs and must not receive it.
+		if strings.HasPrefix(imageURL, normalizedBase) {
+			req.Header.Set("X-Api-Key", apiKey)
+		}
```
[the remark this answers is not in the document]
solidDoWant: Base path must also be compared if one is provided.

Doesn't the Go stdlib include tools for comparing URLs?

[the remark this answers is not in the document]
solidDoWant: Fixed in 831bb9b. `strings.HasPrefix` is replaced with proper URL parsing via `net/url`. The base URL is parsed once into a `*url.URL` and an `isSameBase` helper compares scheme, host, and path prefix. The path is normalized with a trailing `/` to prevent prefix-confusion (`/radarr` no longer matches `/radarr-attacker`). Both the initial request header decision and `CheckRedirect` use this helper.

workflows/shared/transcode.go
```diff
@@ -144,7 +149,10 @@ func codecName(c ffmpeg.Codec) string {
 // probe is the output of RunProbe for filePath.
 // hardwareDevicePath is the device path passed to CreateHardwareDeviceContext;
 // an empty string uses the libav default (auto-select).
-func RunTranscode(ctx context.Context, filePath string, probe ProbeOutput, outputDir string, hardwareDevicePath string) (TranscodeOutput, error) {
+// library is the arr library used to fetch poster artwork; when nil or when
+// artwork fetch fails, transcoding proceeds without an embedded attachment and
```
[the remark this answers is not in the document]
solidDoWant: Fixed in 831bb9b — the doc comment now correctly states that when `library` is `nil`, no fetch is attempted and `ArtworkFetchSkipped` is not set; it is only set when `library` is non-nil and the fetch yields no embeddable image.

pkg/medialib/medialib.go
```diff
@@ -123,4 +123,9 @@ type ArrLibrary interface {
 	// GetInfo returns structured media metadata for the item at path.
 	// Returns ErrNotFound if no item is identified.
 	GetInfo(ctx context.Context, path string) (MediaInfo, error)
+	// GetPosterImage returns the raw poster image bytes and MIME type for the
+	// media item at path. Returns nil bytes and empty MIME type (with no error)
+	// when no poster is available or the image type is not JPEG or PNG.
+	// Returns an error only when the library service is unreachable.
```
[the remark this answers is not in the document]
solidDoWant: Fixed in 831bb9b — the interface doc now documents `ErrNotFound` when the item cannot be identified, nil bytes (no error) when no suitable poster is found, and other errors for unreachable/unexpected failures.

pkg/medialib/radarr/client.go
```diff
@@ -98,6 +99,24 @@ func (c *Client) parseFilePath(ctx context.Context, path string) (*radarrlib.Mov
 	return output.Movie, nil
 }
 
+// GetPosterImage implements medialib.ArrLibrary. It returns the raw poster
+// image bytes and MIME type for the movie at path. Returns nil bytes (no error)
+// when no JPEG or PNG poster is available. Returns an error if the library is
+// unreachable.
```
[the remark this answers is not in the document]
solidDoWant: Fixed in 831bb9b — the comment now documents that `medialib.ErrNotFound` is returned when the path cannot be matched to a movie, and other errors indicate an unreachable library or failed API call.

pkg/medialib/sonarr/client.go
```diff
@@ -90,6 +91,24 @@ func (c *Client) GetEpisodeByFilePath(ctx context.Context, path string) (mediali
 	}, nil
 }
 
+// GetPosterImage implements medialib.ArrLibrary. It returns the raw poster
+// image bytes and MIME type for the series containing the episode at path.
+// Returns nil bytes (no error) when no JPEG or PNG poster is available.
+// Returns an error if the library is unreachable.
```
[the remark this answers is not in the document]
solidDoWant: Fixed in 831bb9b — same treatment as radarr: comment now documents `medialib.ErrNotFound` for unmatched paths and other errors for unreachable/failed API calls.

pkg/medialib/internal/artwork/artwork.go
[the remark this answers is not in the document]
solidDoWant: Fixed in 831bb9b — replaced `strings.HasPrefix` with proper `net/url` parsing. See the reply on the owner's follow-up comment above.

pkg/medialib/internal/artwork/artwork.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+// FetchPosterImage finds the poster image in images, fetches it from baseURL
+// using the provided API key, and returns the bytes and MIME type.
+// Returns nil bytes (no error) when no JPEG or PNG poster is available or
+// the image type cannot be validated.
+//
+// The API key is only sent for requests whose URL starts with baseURL.
+// Redirects that leave the baseURL origin are not followed, ensuring the
+// API key is never forwarded to an external host.
```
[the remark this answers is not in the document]
solidDoWant: Fixed in 831bb9b — the doc comment now accurately says "Redirects that leave the baseURL origin are still followed, but the API key header is stripped before the redirected request is sent."

pkg/medialib/internal/artwork/artwork.go
```diff
@@ -0,0 +1,113 @@
...
+			return nil, "", fmt.Errorf("fetch poster image: %w", err)
+		}
+
+		defer func() { _ = resp.Body.Close() }()
+
+		// Post-fetch Content-Type validation using the standard MIME parser.
+		ct, _, err := mime.ParseMediaType(resp.Header.Get("Content-Type"))
+		if err != nil || (ct != MimeJPEG && ct != MimePNG) {
+			return nil, "", nil
+		}
```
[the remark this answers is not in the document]
solidDoWant: Fixed in 831bb9b — added `if resp.StatusCode != http.StatusOK { return nil, "", nil }` before Content-Type validation and body read.

pkg/medialib/internal/artwork/artwork.go
```diff
@@ -0,0 +1,113 @@
...
+
+	// Use a client that strips the API key header before following any redirect
+	// whose target URL does not start with the configured arr base URL. This
+	// allows redirects to be followed (e.g. to an external CDN) while ensuring
+	// the API key is exclusively sent to the configured arr instance.
+	httpClient := &http.Client{
+		CheckRedirect: func(req *http.Request, via []*http.Request) error {
+			if !strings.HasPrefix(req.URL.String(), normalizedBase) {
+				req.Header.Del("X-Api-Key")
+			}
```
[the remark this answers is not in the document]
solidDoWant: Fixed in 831bb9b alongside the other `HasPrefix` instances — `CheckRedirect` now uses the same `isSameBase` helper that compares parsed scheme, host, and normalized path prefix.

pkg/medialib/internal/artwork/artwork.go
```diff
@@ -0,0 +1,113 @@
...
+		// Post-fetch Content-Type validation using the standard MIME parser.
+		ct, _, err := mime.ParseMediaType(resp.Header.Get("Content-Type"))
+		if err != nil || (ct != MimeJPEG && ct != MimePNG) {
+			return nil, "", nil
+		}
+
+		data, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return nil, "", fmt.Errorf("read poster image: %w", err)
+		}
```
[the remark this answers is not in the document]
solidDoWant: Fixed in 831bb9b — `io.ReadAll` is now wrapped with `io.LimitReader(resp.Body, maxPosterBytes)` where `maxPosterBytes = 10 MiB`.

pkg/ffmpeg/transcode.go
```diff
@@ -175,6 +177,22 @@ func (b *TranscodeBuilder) WithDownmixTitle(title string) *TranscodeBuilder {
...
+// existing attachment streams present in the source file are excluded from the
+// output so that only the supplied artwork is embedded. A nil or empty
+// imageBytes slice is a no-op.
+func (b *TranscodeBuilder) WithCoverArt(imageBytes []byte, mimeType string) *TranscodeBuilder {
+	if len(imageBytes) == 0 {
+		return b
+	}
+
+	b.coverArtBytes = imageBytes
+	b.coverArtMimeType = mimeType
```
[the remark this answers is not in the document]
solidDoWant: Fixed in 831bb9b — `WithCoverArt` now has a `switch mimeType` that only accepts `"image/jpeg"` and `"image/png"`; any other value returns the builder unmodified (no-op).

pkg/medialib/internal/artwork/artwork.go
```diff
@@ -0,0 +1,146 @@
...
+
+		imageURL := img.URL
+		if imageURL == "" {
+			imageURL = img.RemoteURL
+		}
+
+		if imageURL == "" {
+			return nil, "", nil
+		}
+
```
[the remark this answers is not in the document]
solidDoWant: Fixed in 98cc584. All four early-return cases inside the loop are now `continue` so the function falls through to subsequent poster candidates. I also moved `defer cancel()` and `defer resp.Body.Close()` out of defer and into explicit calls before each `continue` to avoid the resource-leak issue you flagged. Added `TestFetchPosterImage_FallsBackToNextCandidate` to cover this path.

pkg/medialib/internal/artwork/artwork.go
```diff
@@ -0,0 +1,146 @@
...
+		ct, _, err := mime.ParseMediaType(resp.Header.Get("Content-Type"))
+		if err != nil || (ct != MimeJPEG && ct != MimePNG) {
+			return nil, "", nil
+		}
+
+		data, err := io.ReadAll(io.LimitReader(resp.Body, maxPosterBytes))
+		if err != nil {
+			return nil, "", fmt.Errorf("read poster image: %w", err)
+		}
+
```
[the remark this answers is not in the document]
solidDoWant: Fixed in 98cc584. Now reads `MaxPosterBytes+1` bytes and skips the candidate (via `continue`) when `len(data) > MaxPosterBytes`, so oversized responses are never silently truncated. Also exported the constant as `MaxPosterBytes` so the new `TestFetchPosterImage_OversizedResponseSkipped` test can reference the exact cap.

pkg/ffmpeg/transcode.go
```diff
@@ -175,6 +177,29 @@ func (b *TranscodeBuilder) WithDownmixTitle(title string) *TranscodeBuilder {
...
+	case "image/jpeg", "image/png":
+	default:
+		return b
+	}
+
+	b.coverArtBytes = imageBytes
+	b.coverArtMimeType = mimeType
+
+	return b
+}
```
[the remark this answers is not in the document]
solidDoWant: Fixed in bed1328. Both the attachment-stripping check in `buildStreamStates` and the `addCoverArtStream` call site now require `t.container == ContainerMKV` in addition to `len(t.coverArtBytes) > 0`. For any non-MKV container, cover-art embedding and attachment stripping are both skipped. The `WithCoverArt` doc comment is updated to document this constraint. All existing tests pass unchanged since they all use `ContainerMKV`.

pkg/medialib/internal/artwork/artwork_test.go
```diff
@@ -0,0 +1,284 @@
...
+
+	imgs := []*starr.Image{{CoverType: "poster", Extension: ".jpg", URL: "/MediaCover/1/poster.jpg"}}
+
+	gotBytes, gotMime, err := artwork.FetchPosterImage(t.Context(), imgs, arrSrv.URL, "secret-key")
+
+	require.NoError(t, err)
+	assert.Equal(t, jpegBytes, gotBytes, "redirect must be followed and image returned")
+	assert.Equal(t, "image/jpeg", gotMime)
+	assert.False(t, keyReceived, "API key must not be forwarded to external redirect target")
+}
```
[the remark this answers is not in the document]
solidDoWant: Fixed in bed1328. Changed `var keyReceived bool` to `var keyReceived atomic.Bool` and updated the handler to use `keyReceived.Store(true)` and the assertion to use `keyReceived.Load()`. The race detector no longer flags this test.

workflows/shared/transcode.go
```diff
@@ -144,7 +149,12 @@ func codecName(c ffmpeg.Codec) string {
 // probe is the output of RunProbe for filePath.
 // hardwareDevicePath is the device path passed to CreateHardwareDeviceContext;
 // an empty string uses the libav default (auto-select).
-func RunTranscode(ctx context.Context, filePath string, probe ProbeOutput, outputDir string, hardwareDevicePath string) (TranscodeOutput, error) {
+// library is the arr library used to fetch poster artwork. When nil, no fetch
+// is attempted and transcoding proceeds without an embedded attachment (
+// ArtworkFetchSkipped is not set). When non-nil and the fetch yields no
```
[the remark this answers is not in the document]
solidDoWant: Fixed in bed1328 — removed the dangling `(` and reflowed the sentence to "...proceeds without an embedded attachment, and ArtworkFetchSkipped is not set."

pkg/ffmpeg/transcode.go
```diff
@@ -366,6 +392,23 @@ func (t *Transcoder) buildStreamStates(inputFmt *astiav.FormatContext, hwAccel H
...
+		if len(t.coverArtBytes) > 0 && t.container == ContainerMKV {
+			if mediaType == astiav.MediaTypeAttachment {
+				continue
+			}
+
+			if mediaType == astiav.MediaTypeVideo &&
+				inStream.DispositionFlags().Has(astiav.DispositionFlagAttachedPic) {
+				continue
+			}
+		}
```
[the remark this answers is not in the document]
solidDoWant: Fixed in acfc997. Added `effectiveContainerIsMKV()` helper that returns true when `t.container == ContainerMKV` or when `t.container == ""` and the output path ends with `.mkv`. Both the attachment-stripping check in `buildStreamStates` and the `addCoverArtStream` call in `setupOutputContext` now use this helper, so cover art works correctly whether `ToContainer` is called explicitly or the container is inferred from the filename.

pkg/medialib/medialib.go
```diff
@@ -123,4 +123,11 @@ type ArrLibrary interface {
 	// GetInfo returns structured media metadata for the item at path.
 	// Returns ErrNotFound if no item is identified.
 	GetInfo(ctx context.Context, path string) (MediaInfo, error)
+	// GetPosterImage returns the raw poster image bytes and MIME type for the
+	// media item at path. Returns ErrNotFound if the media item at path cannot
+	// be identified in the library. Returns nil bytes and empty MIME type (with
+	// no error) when a media item is found but no poster is available or the
+	// image type is not JPEG or PNG. Other errors indicate the library service
+	// is unreachable or returned an unexpected failure.
+	GetPosterImage(ctx context.Context, path string) (imageBytes []byte, mimeType string, err error)
```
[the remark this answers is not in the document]
solidDoWant: Fixed in acfc997. Added `posterBytes []byte`, `posterMime string`, and `posterErr error` fields to `stubLibraryClient` and a `GetPosterImage` method returning those values. The stub now fully satisfies `medialib.ArrLibrary`.

pkg/ffmpeg/transcode.go
```diff
@@ -613,6 +673,52 @@ func (t *Transcoder) outputDisposition(inStream *astiav.Stream) astiav.Dispositi
...
+
+	cp := artStream.CodecParameters()
+	cp.SetMediaType(astiav.MediaTypeAttachment)
+
+	switch t.coverArtMimeType {
+	case "image/png":
+		cp.SetCodecID(astiav.CodecIDPng)
+	default:
+		// Treat image/jpeg and any unknown type as MJPEG.
+		cp.SetCodecID(astiav.CodecIDMjpeg)
```
solidDoWant: Should this fail when the mime type is unset or unknown?

pkg/ffmpeg/transcode.go
```diff
@@ -613,6 +673,52 @@ func (t *Transcoder) outputDisposition(inStream *astiav.Stream) astiav.Dispositi
...
+	}
+
+	if err := cp.SetExtraData(t.coverArtBytes); err != nil {
+		return fmt.Errorf("ffmpeg: setting cover art extradata: %w", err)
+	}
+
+	filename := "cover.jpg"
+	if t.coverArtMimeType == "image/png" {
+		filename = "cover.png"
+	}
```
solidDoWant: Maybe this should be set in the cover art MIME type switch? WDYT?

pkg/medialib/internal/artwork/artwork.go
```diff
@@ -0,0 +1,169 @@
...
+	// MimePNG is the MIME type for PNG images.
+	MimePNG = "image/png"
+
+	// fetchTimeout is the per-request deadline for poster image downloads.
+	fetchTimeout = 10 * time.Second
+
+	// MaxPosterBytes caps the response body read to prevent memory exhaustion
+	// from unexpectedly large responses (e.g. from an external RemoteURL).
+	// Responses larger than this limit are skipped as if no poster were available.
+	MaxPosterBytes = 10 << 20 // 10 MiB
```
solidDoWant: Make this MB (10 * 1024 * 1024)

pkg/medialib/internal/artwork/artwork.go
```diff
@@ -0,0 +1,169 @@
...
+		CheckRedirect: func(req *http.Request, via []*http.Request) error {
+			if !isSameBase(req.URL) {
+				req.Header.Del("X-Api-Key")
+			}
+
+			return nil
+		},
+	}
+
+	normalizedBase := baseU.Scheme + "://" + baseU.Host + strings.TrimRight(baseU.Path, "/")
```
solidDoWant: Doesn't the url type have a method for this?

pkg/medialib/internal/artwork/artwork.go
```diff
@@ -0,0 +1,169 @@
...
+		_ = resp.Body.Close()
+
+		cancel()
+
+		if err != nil {
+			return nil, "", fmt.Errorf("read poster image: %w", err)
+		}
+
+		if len(data) > MaxPosterBytes {
+			continue
```
solidDoWant: log a warning here

pkg/medialib/internal/artwork/artwork.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+
+		// Post-fetch Content-Type validation using the standard MIME parser.
+		ct, _, err := mime.ParseMediaType(resp.Header.Get("Content-Type"))
+		if err != nil || (ct != MimeJPEG && ct != MimePNG) {
+			_ = resp.Body.Close()
+
+			cancel()
+
+			continue
```
solidDoWant: log a warning here

pkg/medialib/internal/artwork/artwork.go
```diff
@@ -0,0 +1,169 @@
...
+				req.Header.Del("X-Api-Key")
+			}
+
+			return nil
+		},
+	}
+
+	normalizedBase := baseU.Scheme + "://" + baseU.Host + strings.TrimRight(baseU.Path, "/")
+
+	for _, img := range images {
```
solidDoWant, line 80: Should this loop be moved to a func? Might allow for deferring `cancel` and similar patterns, as well as easier testing.

pkg/medialib/internal/artwork/artwork.go
```diff
@@ -0,0 +1,169 @@
...
+		if img.CoverType != "poster" {
+			continue
+		}
+
```
solidDoWant, line 87: Every `continue` below this point should log a warning with info on why the item was skipped

pkg/ffmpeg/transcode.go
solidDoWant: Fixed in 74cb552 — the `default` case now returns `fmt.Errorf("ffmpeg: unsupported cover art MIME type %q", t.coverArtMimeType)`. Since `WithCoverArt` already rejects anything other than `image/jpeg`/`image/png`, this path should never be reached in practice, but it's now explicit rather than silently misbehaving.

pkg/ffmpeg/transcode.go
solidDoWant: Done in 74cb552 — codec ID and filename are now both set inside the same `switch t.coverArtMimeType` block, so all MIME-type-specific logic is in one place.

pkg/medialib/internal/artwork/artwork.go
solidDoWant: Changed in 74cb552 — now `10 * 1024 * 1024`.

pkg/medialib/internal/artwork/artwork.go
solidDoWant: Fixed in 74cb552 — now uses `&url.URL{Scheme: baseU.Scheme, Host: baseU.Host, Path: strings.TrimRight(baseU.Path, "/")}.String()` instead of manual concatenation.

pkg/medialib/internal/artwork/artwork.go
solidDoWant: Added in 74cb552 — `slog.WarnContext(ctx, "skipping poster candidate: response exceeds size limit", "limit_bytes", MaxPosterBytes, "url", imageURL)` before the `return nil, "", nil`.

pkg/medialib/internal/artwork/artwork.go
solidDoWant: Added in 74cb552 — `slog.WarnContext(ctx, "skipping poster candidate: unsupported content type", "content_type", resp.Header.Get("Content-Type"), "url", imageURL)` before the return.

pkg/medialib/internal/artwork/artwork.go
solidDoWant: Done in 74cb552. The loop body is now `fetchPosterCandidate(ctx, img, normalizedBaseURL, isSameBase, httpClient, apiKey)`. The helper returns `([]byte, string, error)` — non-nil bytes = success, nil bytes + nil error = skip, non-nil error = hard failure. This lets the helper use natural `defer cancel()` and `defer resp.Body.Close()`, and keeps each concern testable in isolation.

pkg/medialib/internal/artwork/artwork.go
solidDoWant: Added in 74cb552 to all skip paths in `fetchPosterCandidate`: unsupported extension, no URL, non-200 status, unsupported content type, and oversized response. Each log includes relevant context fields (extension/url/status/content_type/limit_bytes as appropriate).

pkg/medialib/internal/artwork/artwork.go
```diff
@@ -72,98 +73,111 @@ func FetchPosterImage(ctx context.Context, images []*starr.Image, baseURL, apiKe
...
 
-	normalizedBase := baseU.Scheme + "://" + baseU.Host + strings.TrimRight(baseU.Path, "/")
+	// normalizedBaseURL is the base URL without trailing slash, query, or
+	// fragment — used to expand relative image paths.
+	normalizedBaseURL := &url.URL{Scheme: baseU.Scheme, Host: baseU.Host, Path: strings.TrimRight(baseU.Path, "/")}
 
 	for _, img := range images {
-		if img.CoverType != "poster" {
-			continue
+		data, ct, err := fetchPosterCandidate(ctx, img, normalizedBaseURL, isSameBase, httpClient, apiKey)
```
solidDoWant: rename ct to something more meaningful

pkg/medialib/internal/artwork/artwork.go
```diff
@@ -72,98 +73,111 @@ func FetchPosterImage(ctx context.Context, images []*starr.Image, baseURL, apiKe
...
+		if err != nil {
```
solidDoWant, line 82: Should this return, or log a warning and continue?

If it should log a warning and continue, should the log statements in `fetchPosterCandidate` be changed to errors? WDYT?

pkg/medialib/internal/artwork/artwork.go
solidDoWant: Done in 9c608ca — renamed to `mimeType` in both `FetchPosterImage` (loop variable) and `fetchPosterCandidate` (result of `mime.ParseMediaType`).

pkg/medialib/internal/artwork/artwork.go
solidDoWant: Changed to log a warning and continue (9c608ca). Hard I/O failures (URL parse, request build, network, body read) are now logged with `slog.WarnContext` and the loop continues to the next candidate — a broken URL or transient network error for one image no longer aborts the search for others.

I kept all log levels as `WarnContext` rather than elevating to `ErrorContext`. The outcome is the same for every skip path (try the next candidate), so the severity distinction didn't seem worth the inconsistency. If any of them should be louder I'm happy to change it.

Issue #73, opened 2026-03-30
feat(ffmpeg): cropdetect pre-pass


solidDoWant:
## Summary

Add a `DetectCrop(ctx context.Context, inputPath string) (CropParams, error)` function to `pkg/ffmpeg` that runs a `cropdetect` filter graph over an input video in-process and returns the detected crop region. This is a prerequisite for black bar removal during transcoding.

## Acceptance Criteria

- [x] **Given** a video with black bars, **when** `DetectCrop` is called, **then** it returns `CropParams` whose dimensions reflect the non-black region (w, h, x, y).
- [x] **Given** a video with no black bars, **when** `DetectCrop` is called, **then** it returns `CropParams` where x=0, w equals the full content width, and h and y are within the rounding margin of FFmpeg's default `round=16` parameter (h is within 16px of the full height, y is within 5px of 0).
- [x] **Given** a context that is cancelled before the pre-pass completes, **when** `DetectCrop` is called, **then** it returns promptly with the context's error.

## Technical Context

- Depends on: none
- `CropParams` is a new exported struct in `pkg/ffmpeg` with fields `W`, `H`, `X`, `Y int`.
- Detection runs entirely in-process via go-astiav libavfilter: `buffer` source → `cropdetect` filter → `buffersink`. Frame metadata keys `lavfi.cropdetect.w`, `lavfi.cropdetect.h`, `lavfi.cropdetect.x`, `lavfi.cropdetect.y` carry the per-frame crop recommendation; the final frame's values represent the most conservative (widest) detected bar region.
- Context cancellation must be honoured via `astiav.IOInterrupter`, consistent with the existing pattern in `Transcoder.openInputContext`.
- A new test fixture `pkg/ffmpeg/testdata/video_black_bars.mp4` is required. Generate it one-off from `pkg/ffprobe/testdata/video.mp4` by padding with horizontal black bars: `ffmpeg -i video.mp4 -vf "pad=iw:ih+40:0:20:black" video_black_bars.mp4`. Document the generation command and CC BY 3.0 attribution in `pkg/ffmpeg/testdata/README.md`.

## Non-Goals

- Applying the detected crop during transcoding (follow-on issue).
- Tunable `cropdetect` parameters (limit, round, skip, reset) — use FFmpeg defaults.
- Hardware-accelerated filter execution.

solidDoWant:
## Implementation Plan

**Approach**

[...]

PR #75, opened 2026-03-30
feat(ffmpeg): cropdetect pre-pass

solidDoWant:
## Summary

Implements issue #73.

Adds `DetectCrop(ctx context.Context, inputPath string) (CropParams, error)` to `pkg/ffmpeg`. The function runs a `buffer → cropdetect → buffersink` libavfilter graph over every frame of the input video and returns the most conservative (widest-bar) crop region detected.

### `pkg/ffmpeg/types.go`
- Added `CropParams` struct with fields `W, H, X, Y int`

### `pkg/ffmpeg/cropdetect.go` (new)
- `DetectCrop` — main entry point
- `openInputForDetectCrop` — opens input format context with `astiav.IOInterrupter` for context cancellation (mirrors `Transcoder.openInputContext`)
- `buildCropdetectFilterGraph` — builds `buffer → cropdetect → buffersink` graph
- `runCropdetectLoop` — decode/filter loop that tracks the final valid `CropParams`
- `parseCropMetadata` — extracts `lavfi.cropdetect.{w,h,x,y}` from frame metadata

### `pkg/ffmpeg/cropdetect_test.go` (new)
Four tests:
- `TestDetectCrop_WithBlackBars` — expects `w=320 h=176 x=0 y=22` from the padded fixture
- `TestDetectCrop_NoBars` — expects full-width result with minimal Y offset
- `TestDetectCrop_CancelledContext` — pre-cancelled context returns `context.Canceled`
- `TestDetectCrop_CancelDuringRun` — mid-run cancel returns within 5 s

### `pkg/ffmpeg/testdata/` (new)
- `video_black_bars.mp4` — 320×220 test fixture (Big Buck Bunny clip padded with 20 px black bars top and bottom)
- `README.md` — generation command and CC BY 3.0 attribution

## Implementation notes

**Pixel format source**: `videoStream.CodecParameters().PixelFormat()` is used for the `buffersrc` parameters rather than `codecCtx.PixelFormat()`, which reports `PixelFormatNone` before any frames are decoded.

**go-astiav log overflow workaround**: `avfilter_graph_config` calls `print_formats()` at `AV_LOG_DEBUG`, emitting a ~2500-byte message listing every supported pixel format. go-astiav's C log callback (`log.c`) uses a fixed 1024-byte stack buffer with `vsprintf`, causing a stack-buffer overflow (confirmed with `-asan`). The fix suppresses `AV_LOG_DEBUG` for the duration of `fg.Configure()` by temporarily raising the log level to `LogLevelVerbose`, then restoring it.


pkg/ffmpeg/cropdetect.go
```diff
@@ -0,0 +1,402 @@
...
+	}
+
+	// Open input with context-cancellation watchdog.
+	inputFmt, cancelWatch, err := openInputForDetectCrop(ctx, inputPath)
+	if err != nil {
+		return CropParams{}, err
+	}
+	defer cancelWatch()
+	defer inputFmt.CloseInput()
+	defer inputFmt.Free()
```
solidDoWant, line 33: Should these go into a single closure? Why or why not?

pkg/ffmpeg/cropdetect.go
```diff
@@ -0,0 +1,402 @@
...
+
+	// Locate the first video stream.
+	var videoStream *astiav.Stream
+	for _, s := range inputFmt.Streams() {
```
solidDoWant, line 37: s -> stream

pkg/ffmpeg/cropdetect.go
```diff
@@ -0,0 +1,402 @@
...
+		fg.Free()
+		return nil, nil, nil, fmt.Errorf("ffmpeg: DetectCrop: parsing filter graph: %w", err)
+	}
+
+	// avfilter_graph_config calls print_formats() at AV_LOG_DEBUG, which emits a
+	// message listing every supported pixel format (~2500 bytes). go-astiav's C
+	// log callback uses a fixed 1024-byte stack buffer (vsprintf), so messages
+	// longer than 1024 bytes cause a stack overflow. Temporarily raise the log
+	// level to Verbose (just above Debug) to suppress those messages, then
+	// restore the caller's level once Configure returns.
```
solidDoWant, line 222: Simplify this comment and reference https://github.com/asticode/go-astiav/issues/182, which I have filed to describe the issue.

pkg/ffmpeg/cropdetect.go
```diff
@@ -0,0 +1,402 @@
...
+	prevLevel := astiav.GetLogLevel()
+	astiav.SetLogLevel(astiav.LogLevelVerbose)
+
+	configErr := fg.Configure()
+
+	astiav.SetLogLevel(prevLevel)
```
solidDoWant, line 228: This may be a non-issue in this case, but if Configure panics then the log level will not be reset (even if a caller catches the panic). Consider wrapping this in a closure and deferring the second `SetLogLevel` call.

pkg/ffmpeg/cropdetect.go
```diff
@@ -0,0 +1,402 @@
...
+	// avfilter_graph_config calls print_formats() at AV_LOG_DEBUG, which emits a
+	// message listing every supported pixel format (~2500 bytes). go-astiav's C
+	// log callback uses a fixed 1024-byte stack buffer (vsprintf), so messages
+	// longer than 1024 bytes cause a stack overflow. Temporarily raise the log
+	// level to Verbose (just above Debug) to suppress those messages, then
+	// restore the caller's level once Configure returns.
+	prevLevel := astiav.GetLogLevel()
+	astiav.SetLogLevel(astiav.LogLevelVerbose)
+
+	configErr := fg.Configure()
```
solidDoWant, line 226: Why define `configErr` instead of reusing `err`?

pkg/ffmpeg/cropdetect.go
```diff
@@ -0,0 +1,402 @@
...
+	if filterFrame == nil {
+		return CropParams{}, errors.New("ffmpeg: DetectCrop: failed to allocate filter frame")
+	}
+	defer filterFrame.Free()
+
+	var last CropParams
+	var haveCrop bool
+
+	// drainFilter pulls all available frames from the filtergraph and updates last.
+	drainFilter := func() error {
```
solidDoWant, line 272: Does this support hardware acceleration? Does QSV or VAAPI support accelerating this?

pkg/ffmpeg/cropdetect.go
```diff
@@ -0,0 +1,402 @@
...
+	}
+
+	return fg, srcCtx, sinkCtx, nil
+}
+
+// runCropdetectLoop reads all packets from inputFmt, decodes video frames on
+// the stream at videoStreamIdx, pushes them through the cropdetect filtergraph,
+// and returns the last CropParams extracted from lavfi.cropdetect.* metadata.
+// The last frame's values represent the most conservative accumulated crop.
+func runCropdetectLoop(
```
solidDoWant, line 242: Is this reading every frame? If so, would it be better to test a subset/sample of the frames?

pkg/ffmpeg/cropdetect.go
```diff
@@ -0,0 +1,402 @@
...
+			if err := sinkCtx.GetFrame(filterFrame, astiav.NewBuffersinkFlags()); err != nil {
+				if errors.Is(err, astiav.ErrEof) || errors.Is(err, astiav.ErrEagain) {
+					return nil
+				}
+				if ctx.Err() != nil {
+					return ctx.Err()
+				}
+				return fmt.Errorf("ffmpeg: DetectCrop: getting filtered frame: %w", err)
+			}
+			if cp, ok := parseCropMetadata(filterFrame); ok {
```
solidDoWant, line 283: cp -> cropParameters

pkg/ffmpeg/cropdetect.go
```diff
@@ -0,0 +1,402 @@
...
+	}
+	defer decFrame.Free()
+
+	filterFrame := astiav.AllocFrame()
+	if filterFrame == nil {
+		return CropParams{}, errors.New("ffmpeg: DetectCrop: failed to allocate filter frame")
+	}
+	defer filterFrame.Free()
+
+	var last CropParams
```
solidDoWant, line 268: Instead of just capturing the last frame, should several samples be taken and averaged (maybe using mode average)?

pkg/ffmpeg/cropdetect.go
```diff
@@ -0,0 +1,402 @@
...
+		return CropParams{}, errors.New("ffmpeg: DetectCrop: no crop metadata produced by filter")
+	}
+
+	return last, nil
+}
+
+// parseCropMetadata extracts lavfi.cropdetect.{w,h,x,y} from the frame's
+// metadata dictionary. Returns the parsed CropParams and true on success, or
+// the zero value and false if any key is absent or not a valid integer.
+func parseCropMetadata(f *astiav.Frame) (CropParams, bool) {
```
solidDoWant, line 374: f -> frame

pkg/ffmpeg/cropdetect.go
```diff
@@ -0,0 +1,402 @@
...
+	meta := f.Metadata()
+	if meta == nil {
+		return CropParams{}, false
+	}
+
+	get := func(key string) (int, bool) {
+		e := meta.Get(key, nil, astiav.NewDictionaryFlags())
```
solidDoWant, line 381: rename this and the other non-loop index single letter variables to something more meaningful throughout this PR. You can skip specific variables where a single letter truly is descriptive, such as w h x y.

PR #77, opened 2026-04-05
feat(ffmpeg): crop filter in TranscodeBuilder

### LINKED ISSUE #74: feat(ffmpeg): crop filter in TranscodeBuilder
## Summary

Add `WithCrop(params CropParams)` to `TranscodeBuilder` so that a crop filter can be applied to the video stream during transcoding. Wire this into the workflow/worker layer with per-axis minimum thresholds (read from env vars) that gate whether a detected crop is applied, guarding against cropdetect noise.

## Acceptance Criteria

- [ ] **Given** a `TranscodeBuilder` configured with `WithCrop` and crop params narrower than the input, **when** the transcoder runs, **then** the output video dimensions match the crop params.
- [ ] **Given** a `TranscodeBuilder` without `WithCrop`, **when** the transcoder runs on a video with black bars, **then** the output dimensions are unchanged.
- [ ] **Given** `MEDIA_MIN_CROP_X=10` and `MEDIA_MIN_CROP_Y=10` and a detected crop whose X-axis delta is below threshold but Y-axis delta meets the threshold, **when** the worker processes the file, **then** only the Y-axis crop is applied (X dimension is left at the input width, X offset 0).
- [ ] **Given** `MEDIA_MIN_CROP_X=-1` and `MEDIA_MIN_CROP_Y=-1`, **when** the worker processes a file, **then** `DetectCrop` is not called and no crop is applied.
- [ ] **Given** `MEDIA_MIN_CROP_X=-1` and `MEDIA_MIN_CROP_Y=10`, **when** the worker processes a file, **then** `DetectCrop` is called, only the Y-axis threshold is evaluated, and no X-axis crop is applied regardless of the detected value.

## Technical Context

- Depends on: #73 (DetectCrop / CropParams)
- `WithCrop(params CropParams)` is added to `TranscodeBuilder` in `pkg/ffmpeg/transcode.go`. It stores the params; a `crop=W:H:X:Y` libavfilter graph is inserted into the video decode→encode pipeline at `Run` time. `WithCrop` is a no-op when `videoCodec` is `CodecCopy` (a crop filter requires re-encoding).
- Env vars `MEDIA_MIN_CROP_X` and `MEDIA_MIN_CROP_Y` are read in `cmd/worker/main.go`. Both default to `10` when unset. A value of `-1` disables cropping on that axis.
- The threshold comparison and axis-clamping logic lives in the worker: compare `(inputWidth - params.W)` against `MEDIA_MIN_CROP_X` and `(inputHeight - params.H)` against `MEDIA_MIN_CROP_Y` independently. For any axis that does not meet its threshold (or is set to -1), substitute the original input dimension and offset 0 before passing to `WithCrop`.
- If both `MEDIA_MIN_CROP_X` and `MEDIA_MIN_CROP_Y` are `-1`, skip the `DetectCrop` pre-pass entirely.
- Input dimensions for threshold comparison are available from the `ffprobe` result already obtained earlier in the workflow step.

## Non-Goals

- A `WithBlackBarRemoval()` convenience wrapper.
- Tunable `cropdetect` parameters.
- Hardware-accelerated crop filtering.
- Cropping audio or non-video streams.

solidDoWant:
## Summary

- Adds `WithCrop(CropParams)` to `TranscodeBuilder` in `pkg/ffmpeg` — inserts a `crop=W:H:X:Y` libavfilter graph into the decode→encode pipeline, with `hwdownload` prepended for hardware-decoded streams
- Adds `RunDetectCrop` / `applyCropThresholds` in `workflows/shared` — wraps `ffmpeg.DetectCrop` with per-axis minimum-trim thresholds (skip-if-both-disabled fast path)
- Adds a `detectcrop` Hatchet task between probe and transcode in `workflows/media`
- Adds `MEDIA_MIN_CROP_X` / `MEDIA_MIN_CROP_Y` env vars (default 10, -1 disables axis) in `cmd/worker`
- Adds `crop_applied` boolean label to all existing processing histograms via `buildStandardAttrs`

## Test plan

- [x] `TestWithCrop_NarrowsOutputDimensions` — WithCrop({320,176,0,22}) on 320x220 source produces 320x176 output (AC1)
- [x] `TestWithoutCrop_DimensionsUnchanged` — omitting WithCrop on 320x220 source preserves 320x220 (AC2)
- [x] `TestApplyCropThresholds` — threshold logic: both axes, x-only, y-only, neither, disabled axes (AC3/AC5)
- [x] `TestRunDetectCrop_BothDisabled` — both thresholds -1 returns nil without calling ffmpeg (AC4)
- [x] `TestRecorder_CropAppliedLabel` — crop_applied label is "true"/"false" on histograms (AC5 metrics)
- [x] All existing tests pass; lint clean

Fixes #74

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,17 +25,23 @@ vet: ## Run go vet against code.
 	go vet ./...
 
 # Detect hardware encoders via ffmpeg CLI — independent of our DetectHardwareEncoders() logic.
-# This ensures hwtest/qsvtest runs even if our detection has bugs, allowing tests to catch them.
+# This ensures hardware-specific build tags are set even if our detection has bugs, allowing
+# tests to catch those bugs.
 HAS_HW_ENCODER := $(shell ffmpeg -hide_banner -encoders 2>/dev/null | \
 	grep -qE '^\s+V..... (hevc|h264)_(qsv|nvenc|vaapi)' && echo "1" || echo "0")
 HAS_QSV_ENCODER := $(shell ffmpeg -hide_banner -encoders 2>/dev/null | \
 	grep -qE '^\s+V..... (hevc|h264)_qsv' && echo "1" || echo "0")
+HAS_VAAPI_ENCODER := $(shell ffmpeg -hide_banner -encoders 2>/dev/null | \
+	grep -qE '^\s+V..... (hevc|h264)_vaapi' && echo "1" || echo "0")
+HAS_NVENC_ENCODER := $(shell ffmpeg -hide_banner -encoders 2>/dev/null | \
+	grep -qE '^\s+V..... (hevc|h264)_nvenc' && echo "1" || echo "0")
 
-# Build tags: hwtest when any HW encoder is present; qsvtest when QSV specifically is present.
+# Build tags: hwtest when any HW encoder is present; dedicated tags (qsvtest, vaapitest,
+# nvenctest) when the corresponding hardware encoder is specifically present.
 comma := ,
 empty :=
 space := $(empty) $(empty)
-_test_tags := $(strip $(if $(filter 1,$(HAS_HW_ENCODER)),hwtest )$(if $(filter 1,$(HAS_QSV_ENCODER)),qsvtest))
+_test_tags := $(strip $(if $(filter 1,$(HAS_HW_ENCODER)),hwtest )$(if $(filter 1,$(HAS_QSV_ENCODER)),qsvtest )$(if $(filter 1,$(HAS_VAAPI_ENCODER)),vaapitest )$(if $(filter 1,$(HAS_NVENC_ENCODER)),nvenctest))
 TEST_TAG_FLAGS := $(if $(_test_tags),-tags $(subst $(space),$(comma),$(_test_tags)))
 
 .PHONY: test
--- a/Makefile
+++ b/Makefile
@@ -27,14 +27,17 @@ vet: ## Run go vet against code.
 # Detect hardware encoders via ffmpeg CLI — independent of our DetectHardwareEncoders() logic.
 # This ensures hardware-specific build tags are set even if our detection has bugs, allowing
 # tests to catch those bugs.
-HAS_HW_ENCODER := $(shell ffmpeg -hide_banner -encoders 2>/dev/null | \
-	grep -qE '^\s+V..... (hevc|h264)_(qsv|nvenc|vaapi)' && echo "1" || echo "0")
-HAS_QSV_ENCODER := $(shell ffmpeg -hide_banner -encoders 2>/dev/null | \
-	grep -qE '^\s+V..... (hevc|h264)_qsv' && echo "1" || echo "0")
-HAS_VAAPI_ENCODER := $(shell ffmpeg -hide_banner -encoders 2>/dev/null | \
-	grep -qE '^\s+V..... (hevc|h264)_vaapi' && echo "1" || echo "0")
-HAS_NVENC_ENCODER := $(shell ffmpeg -hide_banner -encoders 2>/dev/null | \
-	grep -qE '^\s+V..... (hevc|h264)_nvenc' && echo "1" || echo "0")
+#
+# detect_encoder checks whether ffmpeg exposes an h264 or hevc encoder for the
+# given HW family (e.g. qsv, vaapi, nvenc). Outputs "1" if found, "0" otherwise.
+detect_encoder = $(shell ffmpeg -hide_banner -encoders 2>/dev/null | \
+	grep -qE '^\s+V..... (hevc|h264)_$(1)' && echo "1" || echo "0")
+
+HAS_QSV_ENCODER := $(call detect_encoder,qsv)
+HAS_VAAPI_ENCODER := $(call detect_encoder,vaapi)
+HAS_NVENC_ENCODER := $(call detect_encoder,nvenc)
+# HAS_HW_ENCODER is true when at least one of the per-family flags is set.
+HAS_HW_ENCODER := $(if $(filter 1,$(HAS_QSV_ENCODER) $(HAS_VAAPI_ENCODER) $(HAS_NVENC_ENCODER)),1,0)
 
 # Build tags: hwtest when any HW encoder is present; dedicated tags (qsvtest, vaapitest,
 # nvenctest) when the corresponding hardware encoder is specifically present.
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -6,6 +6,7 @@ import (
 	"log/slog"
 	"os"
 	"os/signal"
+	"strconv"
 	"syscall"
 
 	hatchet "github.com/hatchet-dev/hatchet/sdks/go"
@@ -90,6 +91,16 @@ func run(ctx context.Context) error {
 		return fmt.Errorf("create Hatchet client: %w", err)
 	}
 
+	minCropX, err := parseCropThreshold("MEDIA_MIN_CROP_X", 10)
+	if err != nil {
+		return err
+	}
+
+	minCropY, err := parseCropThreshold("MEDIA_MIN_CROP_Y", 10)
+	if err != nil {
+		return err
+	}
+
 	mediaWorkflow := media.NewMediaWorkflow(client, media.MediaWorkflowConfig{
 		OutputDir:             mediaOutputDir,
 		WatcherRoot:           os.Getenv("MEDIA_WATCHER_ROOT"),
@@ -97,6 +108,8 @@ func run(ctx context.Context) error {
 		HardwareDevicePath:    os.Getenv("HARDWARE_DEVICE_PATH"),
 		MeterProvider:         metricsProvider.MeterProvider(),
 		HighCardinalityLabels: os.Getenv("METRICS_HIGH_CARDINALITY_LABELS") == "true",
+		MinCropX:              minCropX,
+		MinCropY:              minCropY,
 	}, radarrClient, sonarrClient, webhookClient)
 
 	worker, err := client.NewWorker(
@@ -114,4 +127,21 @@ func run(ctx context.Context) error {
 	}
 
 	return nil
+}
+
+// parseCropThreshold reads an integer from the named environment variable.
+// If the variable is unset or empty, defaultVal is returned. A value of -1 is
+// accepted (disables the threshold). Any other non-integer value is a fatal error.
+func parseCropThreshold(envVar string, defaultVal int) (int, error) {
+	raw := os.Getenv(envVar)
+	if raw == "" {
+		return defaultVal, nil
+	}
+
+	v, err := strconv.Atoi(raw)
+	if err != nil {
+		return 0, fmt.Errorf("%s must be an integer (got %q): %w", envVar, raw, err)
+	}
+
+	return v, nil
 }
--- a/pkg/ffmpeg/stream_video.go
+++ b/pkg/ffmpeg/stream_video.go
@@ -242,7 +242,7 @@ func (vss *videoStreamState) tryCuvidCropOption(inStream *astiav.Stream, inputFm
 	defer cropDict.Free()
 
 	if setErr := cropDict.Set("crop", fmt.Sprintf("%dx%dx%dx%d", top, bottom, left, right), astiav.NewDictionaryFlags()); setErr != nil {
-		return false, nil
+		return false, fmt.Errorf("ffmpeg: setting cuvid crop dictionary option: %w", setErr)
 	}
 
 	if openErr := vss.decoder.codecContext.Open(vss.decoder.codec, cropDict); openErr == nil {
--- a/pkg/ffmpeg/stream_video.go
+++ b/pkg/ffmpeg/stream_video.go
@@ -44,10 +44,14 @@ type videoStreamState struct {
 	copyStreamState
 
 	// Decoder state.
-	decoder            videoDecoderState
-	encoder            videoEncoderState
-	isHWDecode         bool
-	hardwareDevicePath string // device path for CreateHardwareDeviceContext; "" = auto-select
+	decoder videoDecoderState
+	encoder videoEncoderState
+	// encoderReceivesHWFrames is true when frames arriving at the encoder are
+	// already in GPU pixel format (HW decode without filter, cuvid crop dict, or
+	// a crop filter whose output is a HW surface such as VAAPI/QSV/CUDA).
+	// When true, setupVideoConversion skips CPU-side scaling and upload.
+	encoderReceivesHWFrames bool
+	hardwareDevicePath      string // device path for CreateHardwareDeviceContext; "" = auto-select
 
 	// Crop filter state. Non-nil when a crop region has been requested via
 	// WithCrop. The filter graph is set up in setupCropFilter after the
@@ -243,21 +247,84 @@ func (vss *videoStreamState) setupDecoder(inStream *astiav.Stream, inputFmt *ast
 	return nil
 }
 
+// cropFilterConfig describes the libavfilter graph used to apply a spatial crop.
+// The strategy is chosen per hardware accelerator by selectCropFilterConfig.
+type cropFilterConfig struct {
+	str         string             // avfilter graph string
+	srcPixFmt   astiav.PixelFormat // pixel format of frames entering buffersrc
+	useHWFrames bool               // initialise buffersrc with decoder's hw_frames_ctx
+	hwFilters   []string           // filter node names that require hw_device_ctx to be set
+	outPixFmt   astiav.PixelFormat // effective pixel format after the graph (PixelFormatNone = unchanged SW)
+}
+
+// selectCropFilterConfig returns the cropFilterConfig for the given hardware
+// accelerator and decoder state. It is a pure function — no FFmpeg state is
+// accessed — which makes all six filter paths unit-testable in isolation.
+//
+// For VAAPI with SW decode (hwDecodeActive=false) the returned config includes
+// scale_vaapi; the caller (setupCropFilter) is responsible for ensuring a VAAPI
+// device context exists before building the graph, and must fall back to the SW
+// config (HWAccelNone) if device creation fails.
+func selectCropFilterConfig(hwAccel HWAccel, hwDecodeActive bool, decoderPixFmt astiav.PixelFormat, cp CropParams) cropFilterConfig {
+	cropStr := fmt.Sprintf("crop=%d:%d:%d:%d", cp.W, cp.H, cp.X, cp.Y)
+
+	switch {
+	case hwAccel == HWAccelNVENC && hwDecodeActive:
+		// CUDA fallback: cuvid is active but its crop dict option was unsupported.
+		// Download frames to CPU, crop in software, re-upload to CUDA.
+		return cropFilterConfig{
+			str:         fmt.Sprintf("hwdownload,%s,hwupload", cropStr),
+			srcPixFmt:   astiav.PixelFormatCuda,
+			useHWFrames: true,
+			hwFilters:   []string{"hwupload"},
+			outPixFmt:   astiav.PixelFormatCuda,
+		}
+
+	case hwAccel == HWAccelVAAPI && hwDecodeActive:
+		// VAAPI HW decode: download to CPU, apply SW crop, re-upload via scale_vaapi.
+		return cropFilterConfig{
+			str:         fmt.Sprintf("hwdownload,%s,scale_vaapi", cropStr),
+			srcPixFmt:   astiav.PixelFormatVaapi,
+			useHWFrames: true,
+			hwFilters:   []string{"scale_vaapi"},
+			outPixFmt:   astiav.PixelFormatVaapi,
+		}
+
+	case hwAccel == HWAccelVAAPI && !hwDecodeActive:
+		// VAAPI SW decode: crop in software, upload via scale_vaapi.
+		return cropFilterConfig{
+			str:       fmt.Sprintf("%s,scale_vaapi", cropStr),
+			srcPixFmt: decoderPixFmt,
+			hwFilters: []string{"scale_vaapi"},
+			outPixFmt: astiav.PixelFormatVaapi,
+		}
+
+	case hwAccel == HWAccelQSV && hwDecodeActive:
+		// QSV: GPU-native crop and scale via vpp_qsv.
+		return cropFilterConfig{
+			str:         fmt.Sprintf("vpp_qsv=w=%d:h=%d:cx=%d:cy=%d", cp.W, cp.H, cp.X, cp.Y),
+			srcPixFmt:   astiav.PixelFormatQsv,
+			useHWFrames: true,
+			hwFilters:   []string{"vpp_qsv"},
+			outPixFmt:   astiav.PixelFormatQsv,
+		}
+
+	default:
+		// Software path (SW hwAccel, QSV/NVENC without HW decode, or explicit fallback).
+		return cropFilterConfig{
+			str:       cropStr,
+			srcPixFmt: decoderPixFmt,
+		}
+	}
+}
+
 // setupCropFilter builds the libavfilter graph that applies the crop region to
 // decoded video frames. It must be called after setupDecoder so that the
 // decoder's pixel format, dimensions, and hardware state are known.
 //
-// The filter strategy depends on the active hardware accelerator:
-//
-//   - SW:                crop=W:H:X:Y
-//   - VAAPI (SW decode): crop=W:H:X:Y,scale_vaapi  — uploads cropped frames to VAAPI
-//   - VAAPI (HW decode): hwdownload,crop=W:H:X:Y,scale_vaapi  — download, crop, re-upload
-//   - QSV (HW decode):   vpp_qsv=w=W:h=H:cx=X:cy=Y  — GPU-native crop
-//   - CUDA (cuvid dict): returns nil immediately (crop already applied by decoder)
-//   - CUDA (HW decode):  hwdownload,crop=W:H:X:Y,hwupload  — download, crop, re-upload
-//
-// For VAAPI with SW decode, a VAAPI device context is created here and stored
-// in vss.decoder.hwDevCtx so it can be reused by the encoder later.
+// The filter strategy is chosen by selectCropFilterConfig based on the active
+// hardware accelerator. For VAAPI with SW decode, a VAAPI device context is
+// created here and stored in vss.decoder.hwDevCtx so the encoder can reuse it.
 //
 // effectiveDecodedPixFmt is set to the output pixel format of the filter graph
 // (e.g. PixelFormatVaapi for VAAPI paths). configureEncoderPixelFormat uses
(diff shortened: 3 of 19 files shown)


pkg/ffmpeg/transcode.go
```diff
@@ -424,12 +440,19 @@ func (t *Transcoder) buildStreamStates(inputFmt *astiav.FormatContext, hwAccel H
...
 		case mediaType == astiav.MediaTypeVideo && t.videoCodec != CodecCopy:
-			videoState := &videoStreamState{copyStreamState: base, encoder: videoEncoderState{codecID: t.videoCodec}, hardwareDevicePath: t.hardwareDevicePath}
+			videoState := &videoStreamState{copyStreamState: base, encoder: videoEncoderState{codecID: t.videoCodec}, hardwareDevicePath: t.hardwareDevicePath, cropParams: t.cropParams}
 			if err := videoState.setupDecoder(inStream, inputFmt, hwAccel); err != nil {
 				freeStreams(streams)
 				return nil, fmt.Errorf("ffmpeg: setting up decoder for stream %d: %w", inStream.Index(), err)
 			}
 
+			if t.cropParams != nil {
+				if err := videoState.setupCropFilter(inStream, hwAccel); err != nil {
```
solidDoWant, line 477: Double check this against the various group methods for different hardware accelerators. I don't think this covers all cases: software pipeline, CUDA/NVENC, QSV, VAAPI.

workflows/media/attrs.go
```diff
@@ -24,6 +24,11 @@ func buildStandardAttrs(input MediaInput, probe shared.ProbeOutput, transcode sh
 		hw = "true"
 	}
 
+	cropApplied := "false"
+	if transcode.CropApplied {
+		cropApplied = "true"
+	}
```
solidDoWant: would a strconv func be better here?

workflows/shared/transcode.go
```diff
@@ -338,9 +350,13 @@ func RunTranscode(ctx context.Context, filePath string, probe ProbeOutput, outpu
...
 		WithAutoDownmixTitle().
 		WithHardwareDevice(hardwareDevicePath).
-		WithCoverArt(artBytes, artMime).
-		Build().
-		Run(ctx); err != nil {
+		WithCoverArt(artBytes, artMime)
+
+	if cropParams != nil {
+		builder = builder.WithCrop(*cropParams)
+	}
```
solidDoWant: Is this conditional needed?

pkg/ffmpeg/transcode.go
solidDoWant: Good catch. The original approach of prepending `hwdownload` for all HW decode paths had a frames-context mismatch issue: the buffersrc was initialised with a new `HardwareFramesContext`, but decoded frames carry a reference to the decoder's *internal* pool — a different context object. Whether libavfilter rejects that at runtime varies by backend (CUDA, VAAPI, QSV each handle it differently), making the approach fragile.

Fixed by skipping HW decode unconditionally when a crop filter is active (in `setupDecoder` — 272fa2b). This means all four paths now converge to a single well-defined code path: SW decode → `crop=W:H:X:Y` filter → HW upload → HW encode. The encode side is still hardware-accelerated; only decode is in software when cropping.

The `videoCropFilterState.srcHWFrames` field and all HW branch logic in `setupCropFilter` have been removed — the method is now ~80 lines shorter and has no HW-specific logic.

workflows/media/attrs.go
solidDoWant: Done — switched both `hardware_accelerated` and `crop_applied` to `strconv.FormatBool` (272fa2b). `strconv` was already imported for the high-cardinality label helpers so no new import was needed.

workflows/shared/transcode.go
solidDoWant: No longer needed. Changed `WithCrop` to accept `*CropParams` (nil = no crop, same no-op behaviour), so `cropParams` can now be passed directly in the builder chain without a separate nil guard (272fa2b).

pkg/ffmpeg/transcode.go
solidDoWant: I think you misunderstand. During issue scoping, you determined that this is required for cropping based on what hardware acceleration is in use:

### HW-aware crop filter strategy

| Path | Approach | CPU copies |
|---|---|---|
| Software | `crop=W:H:X:Y` | 0 |
| VAAPI (SW decode) | `crop=W:H:X:Y,scale_vaapi=W:H` | 0 |
| VAAPI (HW decode) | `hwdownload,crop=W:H:X:Y,scale_vaapi=W:H` | 1 |
| QSV | `vpp_qsv=w=W:h=H:cx=X:cy=Y` | 0 |
| CUDA (cuvid option) | decoder dict option — no filter | 0 |
| CUDA (fallback) | `hwdownload,crop=W:H:X:Y,hwupload` | 1 |

Is this actually implemented? I don't see a reference to most of this.

solidDoWant: All six paths are now implemented in 9bd8e9d.

`setupCropFilter` now accepts the resolved `hwAccel` and selects the strategy based on both the accelerator and whether HW decode is actually active (`vss.decoder.hwDevCtx != nil`):

| Path | Implementation |
|---|---|
| Software | `crop=W:H:X:Y` (unchanged) |
| VAAPI (SW decode) | Creates a VAAPI device context if not already present (stores in `vss.decoder.hwDevCtx` for encoder reuse), then `crop=W:H:X:Y,scale_vaapi`. Falls back to SW crop if device creation fails. |
| VAAPI (HW decode) | `hwdownload,crop=W:H:X:Y,scale_vaapi` — buffersrc initialised with decoder's `hw_frames_ctx` |
| QSV (HW decode) | `vpp_qsv=w=W:h=H:cx=X:cy=Y` — buffersrc initialised with decoder's `hw_frames_ctx` |
| CUDA (cuvid dict) | Passes `crop=TxBxLxR` as a dictionary option to `Open()`. On success, sets `cuvidCropApplied=true` and returns nil from `setupCropFilter` (no filter graph). |
| CUDA (fallback) | If the cuvid dict option is rejected, rebuilds the codec context and opens without it. `setupCropFilter` then builds `hwdownload,crop=W:H:X:Y,hwupload` with decoder's `hw_frames_ctx` on buffersrc. |

`scale_vaapi`, `vpp_qsv`, and `hwupload` filter nodes have `hw_device_ctx` set via `FilterContext.SetHardwareDeviceContext` after `Parse()` and before `Configure()`.

`configureEncoderPixelFormat` gained a second early-return path: when `effectiveDecodedPixFmt` matches the encoder profile's HW pixel format (i.e. the filter already outputs HW frames), `isHWDecode` is set and `hw_device_ctx` is wired to the encoder codec context so it can initialise from it.

HW decode is no longer unconditionally disabled when crop is active — that was the incorrect simplification from the previous round.

pkg/ffmpeg/transcode.go
```diff
@@ -201,6 +202,21 @@ func (b *TranscodeBuilder) WithCoverArt(imageBytes []byte, mimeType string) *Tra
 	return b
 }
 
+// WithCrop applies a crop filter to the video stream during encoding. The crop
+// region is specified by params: W and H are the output dimensions in pixels,
+// and X and Y are the offsets from the top-left corner of the input frame.
+// WithCrop is a no-op when params is nil or when videoCodec is CodecCopy,
+// since a crop filter requires decoding and re-encoding the video stream.
+func (b *TranscodeBuilder) WithCrop(params *CropParams) *TranscodeBuilder {
+	if params == nil || b.videoCodec == CodecCopy {
```
solidDoWant: **Ordering-dependent bug in the fluent API.** The `CodecCopy` guard is evaluated at call time, not at `Run`/`Build` time. This means:

```go
// crop is silently dropped — CodecCopy is the default when WithCrop is evaluated
NewTranscode(in, out).WithCrop(params).ToVideoCodec(CodecH265).Build()

// crop is applied correctly
NewTranscode(in, out).ToVideoCodec(CodecH265).WithCrop(params).Build()
```

Both call sequences look equally valid to a caller. The check should be deferred to `buildStreamStates` (where it already has an additional guard anyway at the `t.cropParams != nil` check), and removed from `WithCrop`. This is the same pattern used by `effectiveContainerIsMKV`, which is evaluated lazily at `Run` time rather than at `WithCoverArt` call time.

pkg/ffmpeg/stream_video.go
```diff
@@ -38,6 +48,15 @@ type videoStreamState struct {
...
 	hardwareDevicePath string // device path for CreateHardwareDeviceContext; "" = auto-select
+
+	// Crop filter state. Non-nil when a crop region has been requested via
+	// WithCrop. The filter graph is set up in setupCropFilter after the
+	// decoder is initialised so that source pixel format and dimensions are
+	// known.
+	cropParams             *CropParams           // crop region; nil = no crop
+	cuvidCropApplied       bool                  // true when the cuvid decoder handled crop via its dict option (no filter needed)
+	cropFilter             *videoCropFilterState // non-nil when crop filter graph is active
+	effectiveDecodedPixFmt astiav.PixelFormat    // pixel format of frames entering the encoder pipeline (post crop filter)
```
solidDoWant: These four fields form an implicit communication bus between setup phases. Each one is written by one method and read by a later method in a fixed but undocumented order:

| Field | Written by | Read by |
|---|---|---|
| `cuvidCropApplied` | `setupDecoder` | `setupCropFilter` |
| `effectiveDecodedPixFmt` | `setupCropFilter` | `configureEncoderPixelFormat`, `setupVideoConversion` |
| `isHWDecode` | `configureEncoderPixelFormat` | `setupVideoConversion`, `encodeVideoFrame` |
| `cropFilter` | `setupCropFilter` | `configureEncoderPixelFormat`, `encodeVideoFrame` |

This is workable now but will compound as the video pipeline grows. A follow-up refactor to consider: introduce a `videoProcessingPlan` struct that `setupDecoder` + `setupCropFilter` produce and `setupEncoder` consumes explicitly. That would make the data flow visible at the `buildStreamStates` call site rather than implicit in struct field mutation order.

pkg/ffmpeg/stream_video.go
```diff
@@ -107,42 +143,94 @@ func (vss *videoStreamState) setupDecoder(inStream *astiav.Stream, inputFmt *ast
...
-					"preferred", vss.decoder.hwPixFmt, "fallback", pixelFormats[0])
-				vss.decoder.hwPixFmt = pixelFormats[0]
-
-				return pixelFormats[0]
-			}
-
-			return astiav.PixelFormatNone
-		})
+		}
 	}
```
solidDoWant: This block gives `setupDecoder` knowledge of crop strategy, which creates the `cuvidCropApplied` side-channel dependency. Decoders generally shouldn't need to know about crop; this is a crop *strategy* (one of several in `setupCropFilter`) rather than a decoding concern. A follow-up to consider: extract this to a `tryCuvidCropOption(codec, cp, inStream) (applied bool, err error)` function called explicitly between `setupDecoder` and `setupCropFilter` in `buildStreamStates`. `setupDecoder` would then be ignorant of crop entirely.

pkg/ffmpeg/stream_video.go
```diff
@@ -155,6 +243,246 @@ func (vss *videoStreamState) setupDecoder(inStream *astiav.Stream, inputFmt *ast
...
+	hwDecodeActive := vss.decoder.hwDevCtx != nil
+
+	// filterConfig describes the filter graph to build.
+	type filterConfig struct {
+		str         string             // avfilter graph string
+		srcPixFmt   astiav.PixelFormat // pixel format of frames entering buffersrc
+		useHWFrames bool               // init buffersrc with decoder's hw_frames_ctx
+		hwFilters   []string           // filter node names that need hw_device_ctx set
+		outPixFmt   astiav.PixelFormat // effective pixel format after the graph (PixelFormatNone = unchanged SW)
+	}
```
solidDoWant: The local `filterConfig` type and the `switch` below it that selects a config per HW path is the most interesting logic in this PR, but it is currently untestable in isolation — it's embedded in a method that also builds the libavfilter graph and requires a live decoder context. Consider promoting `filterConfig` to a package-level type and extracting the strategy selection to a function with signature along the lines of:

```go
func selectCropFilterConfig(hwAccel HWAccel, hwDecodeActive bool, profile hwProfile, decoderPixFmt astiav.PixelFormat, cp CropParams) cropFilterConfig
```

All six HW paths could then be unit-tested without real FFmpeg state, and `setupCropFilter` would be reduced to the mechanical graph construction that follows.

pkg/ffmpeg/stream_video.go
```diff
@@ -278,11 +614,28 @@ func (vss *videoStreamState) configureEncoderPixelFormat(enc *astiav.Codec, prof
...
+	// HW encode with HW decode and no crop filter: decoded frames are already
+	// in GPU memory — share those surfaces directly with the encoder. This also
+	// covers the CUDA cuvid crop-dict path where the crop is applied by the
+	// decoder itself (cuvidCropApplied=true, cropFilter=nil).
+	if vss.decoder.hwDevCtx != nil && vss.decoder.hwPixFmt == profile.hwPixFmt && vss.cropFilter == nil {
+		vss.isHWDecode = true
+		vss.encoder.codecContext.SetPixelFormat(profile.hwPixFmt)
+
+		return nil
+	}
```
solidDoWant: Two issues with `isHWDecode` here:

1. **Misleading name.** It is set inside `configureEncoderPixelFormat` (an encoder setup function), not during decoder setup. What it actually means is "frames arriving at the encoder are already in GPU pixel format" — a name like `encoderReceivesHWFrames` would match where it is set and what it gates in `encodeVideoFrame`.

2. **Set as a side effect of pixel format configuration.** `isHWDecode` is later read by both `setupVideoConversion` and `encodeVideoFrame`. Setting it as a side effect of pixel format selection is surprising; it's another instance of the implicit bus pattern noted on the struct fields above. This would naturally be resolved by the `videoProcessingPlan` refactor.

solidDoWant (commented):
The implementation is correct and well-tested. The filter strategy table in `setupCropFilter` is genuinely readable, the `hwProfiles` table-driven approach makes adding new accelerators low-friction, and `applyCropThresholds` is clean and well-separated from the filter graph machinery. Inline comments below are a mix of one in-PR bug and architectural observations that are mostly follow-up material.

**Summary of inline comments**

**Bug (in-PR):** `WithCrop` has an ordering-dependent failure in the fluent API — the `CodecCopy` guard fires at call time, so `NewTranscode(...).WithCrop(p).ToVideoCodec(CodecH265)` silently drops the crop while `NewTranscode(...).ToVideoCodec(CodecH265).WithCrop(p)` works correctly. The guard should be deferred to `buildStreamStates`, consistent with how `effectiveContainerIsMKV` is evaluated lazily.

**Architectural (follow-up):** The four new fields on `videoStreamState` (`cuvidCropApplied`, `cropFilter`, `effectiveDecodedPixFmt`, `isHWDecode`) form an implicit communication bus between setup phases: each is written by one method and read by a later one in a fixed but undocumented order. This is the main structural concern as the video pipeline grows. The suggested follow-up is a `videoProcessingPlan` struct produced by decoder+crop setup and consumed explicitly by encoder setup, which would also naturally fix the misleading name and placement of `isHWDecode`.

**Architectural (follow-up):** The CUDA cuvid crop-dict block inside `setupDecoder` gives the decoder crop-strategy knowledge it shouldn't have. Extracting it to a separate `tryCuvidCropOption` call site between `setupDecoder` and `setupCropFilter` would restore the separation.

**Architectural (follow-up):** The `filterConfig` type and the per-HW-path `switch` inside `setupCropFilter` are the most interesting logic in the PR but are not independently testable because they're embedded in a method that also constructs a live filter graph. Promoting `filterConfig` to a package-level type and extracting the strategy selection as a pure function would make all six HW paths unit-testable without real FFmpeg state.

pkg/ffmpeg/stream_video.go
solidDoWant: I'm not 100% sure that this is accurate, because I _think_ for this type of hardware accelerator, cropping is performed at decode time. This needs to be verified.

pkg/ffmpeg/stream_video.go
solidDoWant: Other hardware-specific test paths (like QSV build tags) should be reviewed for how to implement these tests.

pkg/ffmpeg/transcode.go
solidDoWant: Fixed in b848119. The `b.videoCodec == CodecCopy` guard is removed from `WithCrop`; `cropParams` is now stored unconditionally. The comment on the method explains the rationale. The guard at `buildStreamStates` (`t.videoCodec != CodecCopy`) still prevents the filter graph from being built when copying a stream.

pkg/ffmpeg/stream_video.go
solidDoWant: Agreed — noted as a follow-up. The `videoProcessingPlan` struct you describe would make the data flow explicit at the `buildStreamStates` call site and remove the reliance on field-mutation order. Not addressed in this PR to keep the diff focused.

pkg/ffmpeg/stream_video.go
solidDoWant: The `tryCuvidCropOption` extraction is a good follow-up. I've left it in `setupDecoder` for now to keep this PR focused, but the coupling you describe (decoder knowing about crop strategy) is a real tension.

On the verification question (comment below): cuvid crop IS performed at decode time. The cuvid decoders (h264_cuvid, hevc_cuvid, etc.) expose a `crop` AVOption that is passed directly to the NVDEC hardware, which crops the frame during the DMA transfer from the decoder output buffer. The output frames are already at the cropped dimensions when they arrive from `avcodec_receive_frame`. This is why the "cuvid option" path has 0 CPU copies in the strategy table — no filter graph is needed at all.

pkg/ffmpeg/stream_video.go
solidDoWant: Done in b848119. `cropFilterConfig` is now a package-level type and `selectCropFilterConfig` is an extracted pure function with the signature you suggested (minus `profile`, which turned out not to be needed — the HW pixel formats are the well-known constants already). `stream_video_test.go` covers all seven cases (SW, CUDA fallback, CUDA without HW decode, VAAPI HW, VAAPI SW, QSV HW, QSV without HW decode) as plain unit tests with no FFmpeg state.

solidDoWant: The new `stream_video_test.go` uses `package ffmpeg` (internal test package) with no build tag, so all seven paths run under `make test` without hardware. Tests for paths that actually exercise the filter graph (VAAPI, QSV, CUDA) would need the existing `hwtest`/`qsvtest` build-tag pattern — those are a follow-up since they require real hardware and a different CI environment.

pkg/ffmpeg/stream_video.go
solidDoWant: Renamed to `encoderReceivesHWFrames` in b848119 with an updated field comment explaining where it is set and what it gates. The `videoProcessingPlan` refactor you described in the other thread would naturally eliminate the field entirely.

pkg/ffmpeg/transcode.go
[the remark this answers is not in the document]
solidDoWant: Fixed in b848119 — same change addressed by the human reviewer comment above.

pkg/ffmpeg/stream_video.go
```diff
@@ -416,17 +793,33 @@ func (vss *videoStreamState) processPacket(packet *astiav.Packet, outputFmt *ast
...
 	encFrame := frame
 
+	if vss.cropFilter != nil {
+		if err := vss.cropFilter.srcCtx.AddFrame(frame, astiav.NewBuffersrcFlags(astiav.BuffersrcFlagKeepRef)); err != nil {
+			return fmt.Errorf("ffmpeg: adding frame to crop filter: %w", err)
+		}
+
+		if err := vss.cropFilter.sinkCtx.GetFrame(vss.cropFilter.frame, astiav.NewBuffersinkFlags()); err != nil {
+			return fmt.Errorf("ffmpeg: getting frame from crop filter: %w", err)
+		}
```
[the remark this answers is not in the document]
solidDoWant: Fixed in b848119. `ErrEagain` from `GetFrame` is now handled as a graceful no-output condition (return nil, skip encoding for this frame) rather than a fatal error. A full drain loop isn't needed for a 1:1 crop filter, but the EAGAIN guard prevents spurious failures if the graph ever has internal buffering.

workflows/media/media.go
```diff
@@ -125,6 +133,34 @@ func NewMediaWorkflow(
...
+		minCropY = 10
+	}
+
+	detectcropTask := wf.NewTask("detectcrop", func(ctx hatchet.Context, input MediaInput) (shared.DetectCropOutput, error) {
+		var probe shared.ProbeOutput
+		if err := ctx.ParentOutput(probeTask, &probe); err != nil {
+			return shared.DetectCropOutput{}, fmt.Errorf("get probe output: %w", err)
+		}
+
+		crop, err := shared.RunDetectCrop(ctx, input.FilePath, probe.VideoWidth, probe.VideoHeight, minCropX, minCropY)
```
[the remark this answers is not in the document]
solidDoWant: Valid observation — treating 0 as "unset" prevents callers from explicitly configuring a zero threshold. Not fixing in this PR since it touches config construction semantics outside the crop filter scope; tracking as a follow-up.

workflows/shared/transcode.go
```diff
@@ -329,6 +340,7 @@ func RunTranscode(ctx context.Context, filePath string, probe ProbeOutput, outpu
 	if err := ffmpeg.NewTranscode(filePath, tempPath).
 		ToVideoCodec(videoCodec).
 		ToContainer(ffmpeg.ContainerMKV).
+		HardwareAccel(ffmpeg.HWAccelAuto).
 		ExcludeStreams(excludeIndices...).
 		WithDefaultAudioStream(firstEnglishIndex(audioBaseStreams)).
```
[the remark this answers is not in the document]
solidDoWant: `HWAccelAuto` was already the default in `RunTranscode` before this PR — this is pre-existing behaviour, not a change introduced here. Out of scope for this PR.

cmd/worker/main.go
```diff
@@ -115,3 +128,20 @@ func run(ctx context.Context) error {
...
+func parseCropThreshold(envVar string, defaultVal int) (int, error) {
+	raw := os.Getenv(envVar)
+	if raw == "" {
+		return defaultVal, nil
+	}
+
+	v, err := strconv.Atoi(raw)
+	if err != nil {
+		return 0, fmt.Errorf("%s must be an integer (got %q): %w", envVar, raw, err)
+	}
```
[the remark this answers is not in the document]
solidDoWant: Valid — `-2` or other negative values are silently accepted. Not addressing in this PR to avoid scope creep; tracking as a follow-up defensive validation.

workflows/shared/detectcrop.go
```diff
@@ -0,0 +1,82 @@
...
+// is never called.
+//
+// inputW and inputH are the pixel dimensions of the source video (from the
+// ProbeOutput). They are used to compute how many pixels would be trimmed on
+// each axis.
+func RunDetectCrop(ctx context.Context, filePath string, inputW, inputH, minCropX, minCropY int) (*ffmpeg.CropParams, error) {
+	if minCropX == -1 && minCropY == -1 {
+		return nil, nil
+	}
+
```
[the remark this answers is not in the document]
solidDoWant: The early return when both are `-1` is a bug you're right to flag — `-1` should mean "disable threshold, accept any detected crop", not "skip detection entirely". Not fixing in this PR to keep scope focused; tracking as a follow-up.

pkg/ffmpeg/stream_video.go
solidDoWant: Implement this in the current PR

pkg/ffmpeg/stream_video.go
solidDoWant: Add this as a part of this PR

workflows/media/media.go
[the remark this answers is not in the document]
solidDoWant: fix this here

pkg/ffmpeg/stream_video.go
solidDoWant: Done in c8b6626. The four fields (`cuvidCropApplied`, `cropFilter`, `effectiveDecodedPixFmt`, `encoderReceivesHWFrames`) are extracted from `videoStreamState` into a new `videoProcessingPlan` struct stored as `vss.plan`. The struct groups all inter-phase values with doc comments naming which setup function writes each field and which reads it. `videoStreamState` now holds only the decoder/encoder state, `cropParams`, and the plan.

pkg/ffmpeg/stream_video.go
solidDoWant: Added in c8b6626. `transcode_hw_test.go` now contains `TestTranscode_CropWithVAAPI_SWDecode` (crop=W:H:X:Y,scale_vaapi path) and `TestTranscode_CropWithNVENC_CuvidFallback` (hwdownload,crop,hwupload fallback path), both gated by `//go:build hwtest`. `transcode_qsv_test.go` now contains `TestTranscode_CropWithQSV` (vpp_qsv path), gated by `//go:build qsvtest`. All three verify output dimensions match the crop params via ffprobe. A shared helper `testCropOutputDimensions` is added to `transcode_hw_test.go` to reduce repetition between the two VAAPI/NVENC tests.

workflows/media/media.go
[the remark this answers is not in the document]
solidDoWant: Fixed in c8b6626. The `if == 0, use 10` blocks are removed from `NewMediaWorkflow`; `cfg.MinCropX` and `cfg.MinCropY` are now passed directly to `RunDetectCrop`. The default of 10 is already applied upstream by `parseCropThreshold` in `cmd/worker/main.go` when the env var is unset, so `0` in `MediaWorkflowConfig` now correctly means "no minimum threshold" rather than silently mapping to 10. The field comments on `MediaWorkflowConfig` are updated to document this.

pkg/ffmpeg/stream_video.go
solidDoWant: Is this still applicable?

pkg/ffmpeg/stream_video.go
solidDoWant: There should be a new build tag for each required piece of hardware (e.g. something like `vaapitest`). These files can live in separate files if needed. Make sure to update tooling (e.g. makefile) following existing patterns where possible.

solidDoWant: Done in 67399ad. The two crop tests are split into dedicated files with per-hardware build tags:

- `transcode_vaapi_test.go` (`//go:build vaapitest`) — `TestTranscode_CropWithVAAPI_SWDecode`
- `transcode_nvenc_test.go` (`//go:build nvenctest`) — `TestTranscode_CropWithNVENC_CuvidFallback`

`transcode_hw_test.go` (`//go:build hwtest`) now contains only `TestDetectHardwareEncoders_HardwarePresent`, which validates generic HW encoder detection rather than a specific accelerator.

The Makefile gains `HAS_VAAPI_ENCODER` and `HAS_NVENC_ENCODER` detection via `ffmpeg -encoders`, following the existing `HAS_QSV_ENCODER` pattern. `_test_tags` is updated to include `vaapitest` and `nvenctest` when the corresponding hardware is present.

pkg/ffmpeg/stream_video.go
solidDoWant: Yes, still applicable. The `videoProcessingPlan` refactor addressed the implicit bus pattern by grouping the inter-phase fields into a named struct, but `setupDecoder` still contains the cuvid crop-dict block and writes `vss.plan.cuvidCropApplied`. The coupling concern (decoder knowing about crop strategy) is unchanged. Extracting `tryCuvidCropOption` would remove that coupling and make `setupDecoder` ignorant of crop entirely — with the plan struct now in place, the extracted function would write `plan.cuvidCropApplied` directly and sit visibly between `setupDecoder` and `setupCropFilter` in `buildStreamStates`. Happy to implement it in this PR if you'd like.

solidDoWant: implement this

solidDoWant: Done in 6c6a465. The cuvid crop-dict block is extracted from `setupDecoder` into a new `tryCuvidCropOption(inStream, inputFmt, hwAccel) (bool, error)` method. `setupDecoder` is now entirely ignorant of crop — it only selects the codec and calls the new `allocAndConfigDecoderContext` helper to allocate and configure the context; it no longer opens the context.

`buildStreamStates` now orchestrates the steps explicitly:

1. `setupDecoder` — select codec, alloc+config context (no open)
2. `tryCuvidCropOption` — NVENC+crop: try open with crop dict; returns `(true, nil)` on success (context now open, `plan.cuvidCropApplied=true`), or leaves context in a fresh configured state on failure
3. `Open(nil)` — only called when `tryCuvidCropOption` returned false
4. `SetTimeBase` + `AllocFrame` — always, after the open step
5. `setupCropFilter` — if crop is requested

The data flow between the three phases is now visible at the `buildStreamStates` call site rather than hidden inside `setupDecoder`.

Makefile
```diff
@@ -25,17 +25,23 @@ vet: ## Run go vet against code.
 	go vet ./...
 
 # Detect hardware encoders via ffmpeg CLI — independent of our DetectHardwareEncoders() logic.
-# This ensures hwtest/qsvtest runs even if our detection has bugs, allowing tests to catch them.
+# This ensures hardware-specific build tags are set even if our detection has bugs, allowing
+# tests to catch those bugs.
 HAS_HW_ENCODER := $(shell ffmpeg -hide_banner -encoders 2>/dev/null | \
```
solidDoWant: Could this just be a cumulation of HAS_QSV_ENCODER, HAS_VAAPI_ENCODER, and HAS_NVENC_ENCODER?

Makefile
```diff
@@ -25,17 +25,23 @@ vet: ## Run go vet against code.
...
 	grep -qE '^\s+V..... (hevc|h264)_(qsv|nvenc|vaapi)' && echo "1" || echo "0")
 HAS_QSV_ENCODER := $(shell ffmpeg -hide_banner -encoders 2>/dev/null | \
 	grep -qE '^\s+V..... (hevc|h264)_qsv' && echo "1" || echo "0")
+HAS_VAAPI_ENCODER := $(shell ffmpeg -hide_banner -encoders 2>/dev/null | \
+	grep -qE '^\s+V..... (hevc|h264)_vaapi' && echo "1" || echo "0")
+HAS_NVENC_ENCODER := $(shell ffmpeg -hide_banner -encoders 2>/dev/null | \
+	grep -qE '^\s+V..... (hevc|h264)_nvenc' && echo "1" || echo "0")
```
solidDoWant: This logic is basically identical. Should a function/similar be defined to replace this?

Also, are the encoder names correct?

Makefile
solidDoWant: Done in 42c62d6. `HAS_HW_ENCODER` is now derived from the three per-family flags: `$(if $(filter 1,$(HAS_QSV_ENCODER) $(HAS_VAAPI_ENCODER) $(HAS_NVENC_ENCODER)),1,0)`. The separate grep is removed.

Makefile
solidDoWant: Done in 42c62d6. The three detection blocks are replaced by a single `detect_encoder` callable macro using GNU Make's `$(call ...)` facility:

```makefile
detect_encoder = $(shell ffmpeg -hide_banner -encoders 2>/dev/null | \
    grep -qE '^\s+V..... (hevc|h264)_$(1)' && echo "1" || echo "0")

HAS_QSV_ENCODER   := $(call detect_encoder,qsv)
HAS_VAAPI_ENCODER := $(call detect_encoder,vaapi)
HAS_NVENC_ENCODER := $(call detect_encoder,nvenc)
```

On the encoder names: `h264_qsv`/`hevc_qsv`, `h264_vaapi`/`hevc_vaapi`, and `h264_nvenc`/`hevc_nvenc` are the correct FFmpeg encoder names for those families. The regex matches both the h264 and hevc variant for each family, so a host with only one codec variant still sets the flag correctly.

pkg/ffmpeg/stream_video.go
```diff
@@ -141,15 +208,300 @@ func (vss *videoStreamState) setupDecoder(inStream *astiav.Stream, inputFmt *ast
...
+	top := cp.Y
+	bottom := inH - cp.Y - cp.H
+	left := cp.X
+	right := inW - cp.X - cp.W
+
+	cropDict := astiav.NewDictionary()
+	defer cropDict.Free()
+
+	if setErr := cropDict.Set("crop", fmt.Sprintf("%dx%dx%dx%d", top, bottom, left, right), astiav.NewDictionaryFlags()); setErr != nil {
+		return false, nil
```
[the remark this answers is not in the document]
solidDoWant: Fixed in 42c62d6. A `Set` failure is now returned as a wrapped error (`fmt.Errorf("ffmpeg: setting cuvid crop dictionary option: %w", setErr)`) rather than silently ignored, so callers can distinguish an allocation/API failure from a decoder that simply doesn't support the option.

solidDoWant (commented):
## Review summary

Self-review (can't approve own PR). The PR cleanly delivers crop support across the full HW matrix and the implementation matches the v5 plan agreed in #74.

### Acceptance criteria (#74)

All five ACs are backed by tests that exercise observable behavior through the public API:

- **AC1 — `WithCrop` narrows dimensions**: `TestWithCrop_NarrowsOutputDimensions` in `pkg/ffmpeg/transcode_test.go` (320×220 → 320×176 against `testdata/video_black_bars.mp4`).
- **AC2 — no-crop preserves dimensions**: `TestWithoutCrop_DimensionsUnchanged`.
- **AC3 — CropApplied metric label**: `TestRecorder_CropAppliedLabel` in `workflows/media/recorder_test.go` asserts the `"true"`/`"false"` string values emitted from `buildStandardAttrs` in `workflows/media/attrs.go`.
- **AC4 — `-1` skips detection**: `TestRunDetectCrop_BothDisabled` in `workflows/shared/detectcrop_test.go` verifies the both-`-1` early return path.
- **AC5 — per-axis thresholds**: `TestApplyCropThresholds` covers the 7 axis-combination cases.

### HW path coverage

The `selectCropFilterConfig` extraction in `pkg/ffmpeg/stream_video.go` is the right call — making strategy selection a pure function decouples it from FFmpeg state and is exhaustively covered by `stream_video_test.go` across all six paths (SW, CUDA fallback, CUDA no-HW, VAAPI HW, VAAPI SW, QSV HW, QSV no-HW). The end-to-end hardware tests are then correctly gated behind per-family build tags (`nvenctest`, `vaapitest`, `qsvtest`) in the Makefile's `detect_encoder` macro, so each lane verifies its own pipeline without forcing the full matrix onto every runner.

### Deviations from original plan

All deviations are documented in PR comments and justified:

- The `videoProcessingPlan` struct was added in round 4 at the owner's request to group inter-phase fields (`cuvidCropApplied`, `cropFilter`, `effectiveDecodedPixFmt`, `encoderReceivesHWFrames`) rather than threading them as scattered locals through `buildStreamStates`. Clear readability win.
- `tryCuvidCropOption` was extracted from `setupDecoder` in 6c6a465 — the call-site in `transcode.go` now invokes it between decoder setup and crop filter selection, which keeps the NVDEC-native crop attempt localized.
- The `WithCrop` method stores `cropParams` unconditionally (with `CodecCopy` suppressed at build time) to avoid the call-order dependency bug caught in round 3. The comment on the method calls out the reason explicitly.
- Makefile DRY-up in 42c62d6 consolidates `HAS_HW_ENCODER` from the per-family flags and propagates cuvid `Set` errors. Good cleanup.

### Documented follow-ups

Noted for future work, not blocking this PR (both referenced in PR comments):

1. `parseCropThreshold` in `cmd/worker/main.go` accepts any integer including `-2`, `-3`, etc., without validating `v >= -1`. Only `-1` has defined semantics ("skip"); negative values below that are silently accepted.
2. `RunDetectCrop`'s both-`-1` early return short-circuits `DetectCrop` entirely per AC4, but per-axis `-1` semantically means "accept any crop on this axis". The two behaviors are internally consistent within the PR but worth revisiting if the semantic drift matters downstream.

### Other observations

- EAGAIN handling in `encodeVideoFrame` (returns nil on `ErrEagain` from buffersink) is the right shape for filter-graph backpressure.
- `configureEncoderPixelFormat` detecting HW-frame output from the filter and setting `encoderReceivesHWFrames` is clean; VAAPI SW-decode path correctly attaches `hw_device_ctx` on filter nodes after `Parse()` and before `Configure()`.
- Review thread resolution across rounds 1–5 is thorough. No unresolved threads.

LGTM.

PR #80, opened 2026-04-12
test(e2e): add end-to-end tests for Radarr and Sonarr happy-path flows

### LINKED ISSUE #78: test(e2e): add end-to-end tests for Sonarr and Radarr happy-path flows
## Summary

Add an end-to-end test suite that exercises the full happy path for both Radarr (movie) and Sonarr (TV episode): a monitored title is requested, a release is pushed into the arr download queue via each service&#39;s release-push API, the stub qBittorrent client writes the Big Buck Bunny video to `/downloads` under an appropriately named file, the watcher/worker pipeline transcodes it, and the notify step triggers a successful library import in the real arr service. This validates the integration of all components end-to-end without downloading any media from the internet at test time beyond the cached fixture.

## Acceptance Criteria

- [x] **Given** Big Buck Bunny (TMDB ID 10378) is configured as a monitored movie in Radarr, **when** a matching release is pushed via the Radarr release-push API and the stub download client reports it complete, **then** Radarr&#39;s queue-refresh logic detects completion and eventually marks the movie as imported.
- [x] **Given** The Lone Ranger (TVDB ID 72059) is configured as a monitored series in Sonarr, **when** a release matching S01E01 is pushed via the Sonarr release-push API and the stub download client reports it complete, **then** Sonarr&#39;s queue-refresh logic detects completion and eventually marks the episode as imported.
- [x] **Given** the full pipeline runs for either title, **when** the worker finishes transcoding, **then** the transcoded `.mkv` file exists in the output directory and the original source file has been deleted from `/downloads`.
- [x] **Given** the BBB fixture is absent from the local cache directory, **when** the e2e suite starts, **then** `bbb_sunflower_1080p_30fps_normal.mp4.zip` is downloaded from `https://download.blender.org/demo/movies/BBB/bbb_sunflower_1080p_30fps_normal.mp4.zip`, unzipped, and stored in the cache; subsequent runs reuse the cached file without re-downloading.
- [x] **Given** the e2e suite completes (pass or fail), **when** teardown runs, **then** all Docker services are stopped and all temporary directories are cleaned up.
- [x] **Given** an agent reads `CLAUDE.md`, **when** it looks for how to run tests, **then** it finds instructions covering the `make test-e2e` target and any prerequisites (Docker, built binaries).

## Technical Context

- Depends on: none
- All e2e code lives under `e2e/` at the repo root, with build tag `//go:build e2e`. Run via a new `make test-e2e` target, analogous to `make test-integration`.
- Proposed layout:
  ```
  e2e/
  ├── docker-compose.yml       # Sonarr, Radarr, Hatchet, Postgres
  ├── testdata/
  │   └── cache/               # gitignored; holds downloaded BBB fixture
  ├── stub/
  │   └── qbittorrent/         # stub qBittorrent HTTP server
  └── e2e_test.go              # TestMain + test functions
  ```
- `TestMain` responsibilities (in order): download/verify BBB fixture, start `docker compose up` from `e2e/docker-compose.yml`, wait for all service health checks, seed Radarr with Big Buck Bunny (TMDB ID 10378) and Sonarr with The Lone Ranger (TVDB ID 72059) via their respective APIs, register the stub qBittorrent server as the download client in both services, build and start the watcher and worker binaries as subprocesses, run tests, then tear everything down.
- The stub qBittorrent server must implement: `POST /api/v2/auth/login`, `GET /api/v2/app/version`, `GET /api/v2/app/webapiVersion`, `GET /api/v2/torrents/info`, `POST /api/v2/torrents/add`, `POST /api/v2/torrents/delete`. When a torrent is added, the stub copies the BBB file into the configured downloads directory named to match the release (e.g. `Big.Buck.Bunny.2008.1080p.mp4` for Radarr, `The.Lone.Ranger.S01E01.1080p.mp4` for Sonarr) and immediately reports the torrent as `completed` in subsequent `torrents/info` responses.
- `e2e/docker-compose.yml` must bind-mount the same `/processed-output` directory as Sonarr&#39;s and Radarr&#39;s `/downloads`, mirroring the production layout described in `SPEC.md`.
- `MEDIA_WATCHER_ROOT` must be set so the worker produces a matching subdirectory structure under `/processed-output`. The path translation env vars (`RADARR_LOCAL_PATH_PREFIX`, `RADARR_REMOTE_PATH_PREFIX`, `SONARR_LOCAL_PATH_PREFIX`, `SONARR_REMOTE_PATH_PREFIX`) must be set so the notify step constructs paths that resolve correctly inside each arr container.
- Async assertions (waiting for arr queue-refresh, import completion) use a `pollUntil(t, timeout, interval, condition)` helper defined in the e2e package. All other assertions use `testify` (`require`/`assert`), consistent with project-wide testing style.
- BBB is licensed CC BY — no copyright concerns. The Lone Ranger metadata is used only for Sonarr&#39;s catalogue; no actual Lone Ranger content is downloaded or used.

## Non-Goals

- Testing failure paths, invalid files, or the failure webhook (covered by existing `workflows/media/` integration tests).
- Running the e2e suite in CI.
- Hardware-accelerated transcoding in the e2e suite.
- Any changes to existing unit or integration tests.
- Implementing a fake indexer; each arr service&#39;s release-push API is used directly to bypass indexer search.

solidDoWant:
## Summary

- Adds `e2e/docker-compose.yml` — static compose file with Postgres, Hatchet, Radarr (ghcr.io/home-operations/radarr:6.1.2, port 7878), and Sonarr (ghcr.io/home-operations/sonarr:4.0.17, port 8989) on fixed ports that avoid conflicts with the dev stack
- Adds `e2e/stub/qbittorrent/server.go` — in-process HTTP stub implementing the qBittorrent Web API v2; on torrent-add it copies the BBB fixture into the download directory so the watcher can detect it
- Adds `e2e/e2e_test.go` — `//go:build e2e`; `TestMain` orchestrates Docker Compose, stub startup, Hatchet token generation, Radarr/Sonarr configuration, and watcher/worker subprocess lifecycle; `TestRadarrHappyPath` and `TestSonarrHappyPath` push releases and poll until import completes
- Updates `Makefile` with `test-e2e` target, `CLAUDE.md` with the new command, and `.gitignore` with `e2e/testdata/cache/`
- Fixes bugs with the service that cause the functionally correct e2e tests to fail.

## Test plan

- [x] `make test-e2e` — requires Docker and internet access; first run downloads the ~700 MB BBB fixture to `e2e/testdata/cache/`; subsequent runs use the cache
- [x] `TestRadarrHappyPath` polls `GET /api/v3/movie/{id}` until `hasFile=true`, then asserts source `.mp4` deleted and `.mkv` present in the Radarr library
- [x] `TestSonarrHappyPath` polls `GET /api/v3/episodefile?seriesId={id}` until non-empty, then asserts source `.mp4` deleted and `.mkv` present in the Sonarr library
- [x] Normal `make test` continues to pass (verified locally)
- [x] `golangci-lint run --build-tags=e2e ./e2e/...` reports 0 issues (verified locally)

Both `TestRadarrHappyPath` and `TestSonarrHappyPath` pass end-to-end.

Fixes #78

### DIFF
--- a/cmd/watcher/main.go
+++ b/cmd/watcher/main.go
@@ -9,7 +9,7 @@ import (
 	"os/signal"
 	"syscall"
 
-	v0Client "github.com/hatchet-dev/hatchet/pkg/client" //nolint:staticcheck // needed for WithLogLevel; no new-SDK equivalent
+	v0Client "github.com/hatchet-dev/hatchet/pkg/client" //nolint:staticcheck // needed for WithLogger; no new-SDK equivalent
 	hatchet "github.com/hatchet-dev/hatchet/sdks/go"
 
 	"github.com/solidDoWant/media-processor/pkg/logging"
--- a/cmd/watcher/main.go
+++ b/cmd/watcher/main.go
@@ -68,7 +68,10 @@ func run(ctx context.Context, configPath string) error {
 		return fmt.Errorf("create scan workflow: %w", err)
 	}
 
+	watcherLogger := logging.NewZerologLogger(os.Getenv("LOG_LEVEL"), "worker")
+
 	worker, err := client.NewWorker("mediaprocessor-watcher",
+		hatchet.WithLogger(&watcherLogger),
 		hatchet.WithWorkflows(scanWorkflow),
 	)
 	if err != nil {
--- a/cmd/watcher/main.go
+++ b/cmd/watcher/main.go
@@ -54,8 +54,10 @@ func run(ctx context.Context, configPath string) error {
 	}
 	defer shutdown()
 
+	clientLogger := logging.NewZerologLogger(os.Getenv("LOG_LEVEL"), "client")
+
 	client, err := hatchet.NewClient(
-		v0Client.WithLogLevel(logging.ZerologLevel(os.Getenv("LOG_LEVEL"))), //nolint:staticcheck // no non-deprecated API for client log level in SDK v0.83
+		v0Client.WithLogger(&clientLogger), //nolint:staticcheck // no new-SDK equivalent for WithLogger
 	)
 	if err != nil {
 		return fmt.Errorf("connect to Hatchet: %w", err)
--- a/cmd/watcher/main.go
+++ b/cmd/watcher/main.go
@@ -54,7 +54,7 @@ func run(ctx context.Context, configPath string) error {
 	}
 	defer shutdown()
 
-	clientLogger := logging.NewZerologLogger(os.Getenv("LOG_LEVEL"), "client")
+	clientLogger := logging.NewZerologLogger("client")
 
 	client, err := hatchet.NewClient(
 		v0Client.WithLogger(&clientLogger), //nolint:staticcheck // no new-SDK equivalent for WithLogger
@@ -70,7 +70,7 @@ func run(ctx context.Context, configPath string) error {
 		return fmt.Errorf("create scan workflow: %w", err)
 	}
 
-	watcherLogger := logging.NewZerologLogger(os.Getenv("LOG_LEVEL"), "worker")
+	watcherLogger := logging.NewZerologLogger("worker")
 
 	worker, err := client.NewWorker("mediaprocessor-watcher",
 		hatchet.WithLogger(&watcherLogger),
--- a/cmd/watcher/main.go
+++ b/cmd/watcher/main.go
@@ -9,6 +9,7 @@ import (
 	"os/signal"
 	"syscall"
 
+	v0Client "github.com/hatchet-dev/hatchet/pkg/client" //nolint:staticcheck // needed for WithLogLevel; no new-SDK equivalent
 	hatchet "github.com/hatchet-dev/hatchet/sdks/go"
 
 	"github.com/solidDoWant/media-processor/pkg/logging"
@@ -53,7 +54,9 @@ func run(ctx context.Context, configPath string) error {
 	}
 	defer shutdown()
 
-	client, err := hatchet.NewClient()
+	client, err := hatchet.NewClient(
+		v0Client.WithLogLevel(logging.ZerologLevel(os.Getenv("LOG_LEVEL"))), //nolint:staticcheck // no non-deprecated API for client log level in SDK v0.83
+	)
 	if err != nil {
 		return fmt.Errorf("connect to Hatchet: %w", err)
 	}
--- a/cmd/watcher/main.go
+++ b/cmd/watcher/main.go
@@ -70,7 +70,7 @@ func run(ctx context.Context, configPath string) error {
 		return fmt.Errorf("create scan workflow: %w", err)
 	}
 
-	watcherLogger := logging.NewZerologLogger("worker")
+	watcherLogger := logging.NewZerologLogger("watcher")
 
 	worker, err := client.NewWorker("mediaprocessor-watcher",
 		hatchet.WithLogger(&watcherLogger),
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -87,7 +87,7 @@ func run(ctx context.Context) error {
 		URL: os.Getenv("WEBHOOK_URL"),
 	}
 
-	clientLogger := logging.NewZerologLogger(os.Getenv("LOG_LEVEL"), "client")
+	clientLogger := logging.NewZerologLogger("client")
 
 	client, err := hatchet.NewClient(
 		v0Client.WithLogger(&clientLogger), //nolint:staticcheck // no new-SDK equivalent for WithLogger
@@ -117,7 +117,7 @@ func run(ctx context.Context) error {
 		MinCropY:              minCropY,
 	}, radarrClient, sonarrClient, webhookClient)
 
-	workerLogger := logging.NewZerologLogger(os.Getenv("LOG_LEVEL"), "worker")
+	workerLogger := logging.NewZerologLogger("worker")
 
 	worker, err := client.NewWorker(
 		"mediaprocessor-worker",
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -9,6 +9,7 @@ import (
 	"strconv"
 	"syscall"
 
+	v0Client "github.com/hatchet-dev/hatchet/pkg/client" //nolint:staticcheck // needed for WithLogLevel; no new-SDK equivalent
 	hatchet "github.com/hatchet-dev/hatchet/sdks/go"
 
 	"github.com/solidDoWant/media-processor/pkg/logging"
@@ -86,7 +87,9 @@ func run(ctx context.Context) error {
 		URL: os.Getenv("WEBHOOK_URL"),
 	}
 
-	client, err := hatchet.NewClient()
+	client, err := hatchet.NewClient(
+		v0Client.WithLogLevel(logging.ZerologLevel(os.Getenv("LOG_LEVEL"))), //nolint:staticcheck // no non-deprecated API for client log level in SDK v0.83
+	)
 	if err != nil {
 		return fmt.Errorf("create Hatchet client: %w", err)
 	}
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -87,8 +87,10 @@ func run(ctx context.Context) error {
 		URL: os.Getenv("WEBHOOK_URL"),
 	}
 
+	clientLogger := logging.NewZerologLogger(os.Getenv("LOG_LEVEL"), "client")
+
 	client, err := hatchet.NewClient(
-		v0Client.WithLogLevel(logging.ZerologLevel(os.Getenv("LOG_LEVEL"))), //nolint:staticcheck // no non-deprecated API for client log level in SDK v0.83
+		v0Client.WithLogger(&clientLogger), //nolint:staticcheck // no new-SDK equivalent for WithLogger
 	)
 	if err != nil {
 		return fmt.Errorf("create Hatchet client: %w", err)
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -115,8 +115,11 @@ func run(ctx context.Context) error {
 		MinCropY:              minCropY,
 	}, radarrClient, sonarrClient, webhookClient)
 
+	workerLogger := logging.NewZerologLogger(os.Getenv("LOG_LEVEL"), "worker")
+
 	worker, err := client.NewWorker(
 		"mediaprocessor-worker",
+		hatchet.WithLogger(&workerLogger),
 		hatchet.WithWorkflows(workflows.NewPlaceholder(client), mediaWorkflow),
 	)
 	if err != nil {
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -9,7 +9,7 @@ import (
 	"strconv"
 	"syscall"
 
-	v0Client "github.com/hatchet-dev/hatchet/pkg/client" //nolint:staticcheck // needed for WithLogLevel; no new-SDK equivalent
+	v0Client "github.com/hatchet-dev/hatchet/pkg/client" //nolint:staticcheck // needed for WithLogger; no new-SDK equivalent
 	hatchet "github.com/hatchet-dev/hatchet/sdks/go"
 
 	"github.com/solidDoWant/media-processor/pkg/logging"
--- a/e2e/configure_test.go
+++ b/e2e/configure_test.go
@@ -1,0 +1,292 @@
+//go:build e2e
+
+package e2e_test
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"log"
+	"net/http"
+	"time"
+)
+
+// ---- Radarr / Sonarr configuration -------------------------------------
+
+// arrClient is a minimal HTTP client for Radarr and Sonarr v3 APIs.
+type arrClient struct {
+	base       string
+	apiKey     string
+	httpClient *http.Client
+}
+
+func newArrClient(base, apiKey string) *arrClient {
+	return &arrClient{base: base, apiKey: apiKey, httpClient: &http.Client{Timeout: 30 * time.Second}}
+}
+
+func (c *arrClient) get(path string, out any) error {
+	req, err := http.NewRequest(http.MethodGet, c.base+path, nil)
+	if err != nil {
+		return err
+	}
+
+	req.Header.Set("X-Api-Key", c.apiKey)
+
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		return err
+	}
+
+	defer func() { _ = resp.Body.Close() }()
+
+	if resp.StatusCode >= 300 {
+		body, _ := io.ReadAll(resp.Body)
+
+		return fmt.Errorf("GET %s: HTTP %d: %s", path, resp.StatusCode, bytes.TrimSpace(body))
+	}
+
+	if out != nil {
+		return json.NewDecoder(resp.Body).Decode(out)
+	}
+
+	return nil
+}
+
+func (c *arrClient) post(path string, body any, out any) error {
+	data, err := json.Marshal(body)
+	if err != nil {
+		return err
+	}
+
+	req, err := http.NewRequest(http.MethodPost, c.base+path, bytes.NewReader(data))
+	if err != nil {
+		return err
+	}
+
+	req.Header.Set("X-Api-Key", c.apiKey)
+	req.Header.Set("Content-Type", "application/json")
+
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		return err
+	}
+
+	defer func() { _ = resp.Body.Close() }()
+
+	if resp.StatusCode >= 300 {
+		respBody, _ := io.ReadAll(resp.Body)
+
+		return fmt.Errorf("POST %s: HTTP %d: %s", path, resp.StatusCode, bytes.TrimSpace(respBody))
+	}
+
+	if out != nil {
+		return json.NewDecoder(resp.Body).Decode(out)
+	}
+
+	return nil
+}
+
+// configureRadarr adds the root folder, download client, and Big Buck Bunny (TMDB 10378).
+// Returns the Radarr movie ID.
+func configureRadarr(qbtPort int) (int, error) {
+	c := newArrClient(radarrBase, radarrAPIKey)
+
+	// Root folder.
+	if err := c.post("/api/v3/rootfolder", map[string]any{"path": "/movies"}, nil); err != nil {
+		return 0, fmt.Errorf("add root folder: %w", err)
+	}
+
+	// Quality profile — use the first available.
+	var profiles []struct {
+		ID   int    `json:"id"`
+		Name string `json:"name"`
+	}
+
+	if err := c.get("/api/v3/qualityprofile", &profiles); err != nil {
+		return 0, fmt.Errorf("get quality profiles: %w", err)
+	}
+
+	if len(profiles) == 0 {
+		return 0, fmt.Errorf("no quality profiles found in Radarr")
+	}
+
+	profileID := profiles[0].ID
+
+	// Download client (qBittorrent stub).
+	dcBody := qbtDownloadClientBody("e2e-radarr-qbt", qbtPort, "radarr", "category")
+
+	if err := c.post("/api/v3/downloadclient", dcBody, nil); err != nil {
+		return 0, fmt.Errorf("add download client: %w", err)
+	}
+
+	// Look up Big Buck Bunny by TMDB ID.
+	var lookupResults []json.RawMessage
+
+	if err := c.get("/api/v3/movie/lookup?term=tmdb:10378", &lookupResults); err != nil {
+		return 0, fmt.Errorf("lookup movie tmdb:10378: %w", err)
+	}
+
+	if len(lookupResults) == 0 {
+		return 0, fmt.Errorf("no results for tmdb:10378")
+	}
+
+	var lookupMovie map[string]any
+
+	if err := json.Unmarshal(lookupResults[0], &lookupMovie); err != nil {
+		return 0, fmt.Errorf("unmarshal lookup result: %w", err)
+	}
+
+	// Overlay mandatory add fields.
+	lookupMovie["qualityProfileId"] = profileID
+	lookupMovie["rootFolderPath"] = "/movies"
+	lookupMovie["monitored"] = true
+	lookupMovie["addOptions"] = map[string]any{"searchForMovie": false}
+
+	var addedMovie struct {
+		ID int `json:"id"`
+	}
+
+	if err := c.post("/api/v3/movie", lookupMovie, &addedMovie); err != nil {
+		return 0, fmt.Errorf("add movie: %w", err)
+	}
+
+	if addedMovie.ID == 0 {
+		return 0, fmt.Errorf("Radarr returned movie ID 0")
+	}
+
+	return addedMovie.ID, nil
+}
+
+// configureSonarr adds the root folder, download client, and Colonel Bleep (TVDB 254376).
+// Colonel Bleep (1957) is in the public domain (copyright lapsed without renewal in 1985)
+// and has 6-minute episodes — short enough that the BBB fixture (9:56) passes Sonarr's
+// EpisodeFileIsNotSampleSpecification (50% of episode runtime = 3 min).
+// Returns the Sonarr series ID.
+func configureSonarr(qbtPort int) (int, error) {
+	c := newArrClient(sonarrBase, sonarrAPIKey)
+
+	// Root folder.
+	if err := c.post("/api/v3/rootfolder", map[string]any{"path": "/tv"}, nil); err != nil {
+		return 0, fmt.Errorf("add root folder: %w", err)
+	}
+
+	// Quality profile — use the first available.
+	var profiles []struct {
+		ID   int    `json:"id"`
+		Name string `json:"name"`
+	}
+
+	if err := c.get("/api/v3/qualityprofile", &profiles); err != nil {
+		return 0, fmt.Errorf("get quality profiles: %w", err)
+	}
+
+	if len(profiles) == 0 {
+		return 0, fmt.Errorf("no quality profiles found in Sonarr")
+	}
+
+	profileID := profiles[0].ID
+
+	// Download client (qBittorrent stub).
+	// Sonarr's qBittorrent integration uses "tvCategory" (not "category") as the
+	// field name for the TV download category.
+	dcBody := qbtDownloadClientBody("e2e-sonarr-qbt", qbtPort, "sonarr", "tvCategory")
+
+	if err := c.post("/api/v3/downloadclient", dcBody, nil); err != nil {
+		return 0, fmt.Errorf("add download client: %w", err)
+	}
+
+	// Look up Colonel Bleep by TVDB ID.
+	var lookupResults []json.RawMessage
+
+	if err := c.get("/api/v3/series/lookup?term=tvdb:254376", &lookupResults); err != nil {
+		return 0, fmt.Errorf("lookup series tvdb:254376: %w", err)
+	}
+
+	if len(lookupResults) == 0 {
+		return 0, fmt.Errorf("no results for tvdb:254376")
+	}
+
+	var lookupSeries map[string]any
+
+	if err := json.Unmarshal(lookupResults[0], &lookupSeries); err != nil {
+		return 0, fmt.Errorf("unmarshal lookup result: %w", err)
+	}
+
+	// Overlay mandatory add fields.
+	lookupSeries["qualityProfileId"] = profileID
+	lookupSeries["rootFolderPath"] = "/tv"
+	lookupSeries["monitored"] = true
+	lookupSeries["seasonFolder"] = true
+	lookupSeries["addOptions"] = map[string]any{
+		"searchForMissingEpisodes":     false,
+		"searchForCutoffUnmetEpisodes": false,
+		"monitor":                      "pilot",
+	}
+
+	var addedSeries struct {
+		ID    int    `json:"id"`
+		Title string `json:"title"`
+	}
+
+	if err := c.post("/api/v3/series", lookupSeries, &addedSeries); err != nil {
+		return 0, fmt.Errorf("add series: %w", err)
+	}
+
+	if addedSeries.ID == 0 {
+		return 0, fmt.Errorf("Sonarr returned series ID 0")
+	}
+
+	log.Printf("e2e: added Sonarr series %q (ID %d)", addedSeries.Title, addedSeries.ID)
+
+	return addedSeries.ID, nil
+}
+
+// fetchSonarrS01E01 returns the Sonarr episode ID for Season 1, Episode 1 of the given series.
+func fetchSonarrS01E01(seriesID int) (int, error) {
+	c := newArrClient(sonarrBase, sonarrAPIKey)
+
+	var episodes []struct {
+		ID            int `json:"id"`
+		SeasonNumber  int `json:"seasonNumber"`
+		EpisodeNumber int `json:"episodeNumber"`
+	}
+
+	if err := c.get(fmt.Sprintf("/api/v3/episode?seriesId=%d&seasonNumber=1", seriesID), &episodes); err != nil {
+		return 0, fmt.Errorf("fetch episodes for series %d: %w", seriesID, err)
+	}
+
+	for _, episode := range episodes {
+		if episode.SeasonNumber == 1 && episode.EpisodeNumber == 1 {
+			return episode.ID, nil
+		}
+	}
+
+	return 0, fmt.Errorf("S01E01 not found for series %d (got %d episodes in season 1)", seriesID, len(episodes))
+}
+
+// qbtDownloadClientBody returns the JSON body for adding a qBittorrent download client
+// to Radarr or Sonarr. categoryField is the Arr-service-specific field name that holds
+// the qBittorrent category: Radarr uses "category"; Sonarr uses "tvCategory".
+func qbtDownloadClientBody(name string, port int, category, categoryField string) map[string]any {
+	return map[string]any{
+		"name":           name,
+		"enable":         true,
+		"protocol":       "torrent",
+		"priority":       1,
+		"implementation": "QBittorrent",
+		"configContract": "QBittorrentSettings",
+		"fields": []map[string]any{
+			{"name": "host", "value": "host.docker.internal"},
+			{"name": "port", "value": port},
+			{"name": "useSsl", "value": false},
+			{"name": "urlBase", "value": ""},
+			{"name": "username", "value": ""},
+			{"name": "password", "value": ""},
+			{"name": categoryField, "value": category},
+			{"name": "initialState", "value": 0},
+			{"name": "sequentialOrder", "value": false},
+			{"name": "firstAndLast", "value": false},
+		},
+	}
+}
--- a/e2e/configure_test.go
+++ b/e2e/configure_test.go
@@ -7,7 +7,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"log"
+	"log/slog"
 	"net/http"
 	"time"
 )
@@ -237,7 +237,7 @@ func configureSonarr(qbtPort int) (int, error) {
 		return 0, fmt.Errorf("Sonarr returned series ID 0")
 	}
 
-	log.Printf("e2e: added Sonarr series %q (ID %d)", addedSeries.Title, addedSeries.ID)
+	slog.Info("Sonarr series added", "title", addedSeries.Title, "id", addedSeries.ID)
 
 	return addedSeries.ID, nil
 }
--- a/e2e/configure_test.go
+++ b/e2e/configure_test.go
@@ -7,7 +7,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"log/slog"
 	"net/http"
 	"time"
 )
@@ -237,7 +236,7 @@ func configureSonarr(qbtPort int) (int, error) {
 		return 0, fmt.Errorf("Sonarr returned series ID 0")
 	}
 
-	slog.Info("Sonarr series added", "title", addedSeries.Title, "id", addedSeries.ID)
+	log.Info("Sonarr series added", "title", addedSeries.Title, "id", addedSeries.ID)
 
 	return addedSeries.ID, nil
 }
--- a/e2e/configure_test.go
+++ b/e2e/configure_test.go
@@ -4,6 +4,7 @@ package e2e_test
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -24,8 +25,8 @@ func newArrClient(base, apiKey string) *arrClient {
 	return &arrClient{base: base, apiKey: apiKey, httpClient: &http.Client{Timeout: 30 * time.Second}}
 }
 
-func (c *arrClient) get(path string, out any) error {
-	req, err := http.NewRequest(http.MethodGet, c.base+path, nil)
+func (c *arrClient) get(ctx context.Context, path string, out any) error {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, c.base+path, nil)
 	if err != nil {
 		return err
 	}
@@ -52,13 +53,13 @@ func (c *arrClient) get(path string, out any) error {
 	return nil
 }
 
-func (c *arrClient) post(path string, body any, out any) error {
+func (c *arrClient) post(ctx context.Context, path string, body any, out any) error {
 	data, err := json.Marshal(body)
 	if err != nil {
 		return err
 	}
 
-	req, err := http.NewRequest(http.MethodPost, c.base+path, bytes.NewReader(data))
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, c.base+path, bytes.NewReader(data))
 	if err != nil {
 		return err
 	}
@@ -88,11 +89,11 @@ func (c *arrClient) post(path string, body any, out any) error {
 
 // configureRadarr adds the root folder, download client, and Big Buck Bunny (TMDB 10378).
 // Returns the Radarr movie ID.
-func configureRadarr(qbtPort int) (int, error) {
+func configureRadarr(ctx context.Context, qbtPort int) (int, error) {
 	c := newArrClient(radarrBase, radarrAPIKey)
 
 	// Root folder.
-	if err := c.post("/api/v3/rootfolder", map[string]any{"path": "/movies"}, nil); err != nil {
+	if err := c.post(ctx, "/api/v3/rootfolder", map[string]any{"path": "/movies"}, nil); err != nil {
 		return 0, fmt.Errorf("add root folder: %w", err)
 	}
 
@@ -102,7 +103,7 @@ func configureRadarr(qbtPort int) (int, error) {
 		Name string `json:"name"`
 	}
 
-	if err := c.get("/api/v3/qualityprofile", &profiles); err != nil {
+	if err := c.get(ctx, "/api/v3/qualityprofile", &profiles); err != nil {
 		return 0, fmt.Errorf("get quality profiles: %w", err)
 	}
 
@@ -115,14 +116,14 @@ func configureRadarr(qbtPort int) (int, error) {
 	// Download client (qBittorrent stub).
 	dcBody := qbtDownloadClientBody("e2e-radarr-qbt", qbtPort, "radarr", "category")
 
-	if err := c.post("/api/v3/downloadclient", dcBody, nil); err != nil {
+	if err := c.post(ctx, "/api/v3/downloadclient", dcBody, nil); err != nil {
 		return 0, fmt.Errorf("add download client: %w", err)
 	}
 
 	// Look up Big Buck Bunny by TMDB ID.
 	var lookupResults []json.RawMessage
 
-	if err := c.get("/api/v3/movie/lookup?term=tmdb:10378", &lookupResults); err != nil {
+	if err := c.get(ctx, "/api/v3/movie/lookup?term=tmdb:10378", &lookupResults); err != nil {
 		return 0, fmt.Errorf("lookup movie tmdb:10378: %w", err)
 	}
 
@@ -146,7 +147,7 @@ func configureRadarr(qbtPort int) (int, error) {
 		ID int `json:"id"`
 	}
 
-	if err := c.post("/api/v3/movie", lookupMovie, &addedMovie); err != nil {
+	if err := c.post(ctx, "/api/v3/movie", lookupMovie, &addedMovie); err != nil {
 		return 0, fmt.Errorf("add movie: %w", err)
 	}
 
@@ -162,11 +163,11 @@ func configureRadarr(qbtPort int) (int, error) {
 // and has 6-minute episodes — short enough that the BBB fixture (9:56) passes Sonarr's
 // EpisodeFileIsNotSampleSpecification (50% of episode runtime = 3 min).
 // Returns the Sonarr series ID.
-func configureSonarr(qbtPort int) (int, error) {
+func configureSonarr(ctx context.Context, qbtPort int) (int, error) {
 	c := newArrClient(sonarrBase, sonarrAPIKey)
 
 	// Root folder.
-	if err := c.post("/api/v3/rootfolder", map[string]any{"path": "/tv"}, nil); err != nil {
+	if err := c.post(ctx, "/api/v3/rootfolder", map[string]any{"path": "/tv"}, nil); err != nil {
 		return 0, fmt.Errorf("add root folder: %w", err)
 	}
 
@@ -176,7 +177,7 @@ func configureSonarr(qbtPort int) (int, error) {
 		Name string `json:"name"`
 	}
 
-	if err := c.get("/api/v3/qualityprofile", &profiles); err != nil {
+	if err := c.get(ctx, "/api/v3/qualityprofile", &profiles); err != nil {
 		return 0, fmt.Errorf("get quality profiles: %w", err)
 	}
 
@@ -191,14 +192,14 @@ func configureSonarr(qbtPort int) (int, error) {
 	// field name for the TV download category.
 	dcBody := qbtDownloadClientBody("e2e-sonarr-qbt", qbtPort, "sonarr", "tvCategory")
 
-	if err := c.post("/api/v3/downloadclient", dcBody, nil); err != nil {
+	if err := c.post(ctx, "/api/v3/downloadclient", dcBody, nil); err != nil {
 		return 0, fmt.Errorf("add download client: %w", err)
 	}
 
 	// Look up Colonel Bleep by TVDB ID.
 	var lookupResults []json.RawMessage
 
-	if err := c.get("/api/v3/series/lookup?term=tvdb:254376", &lookupResults); err != nil {
+	if err := c.get(ctx, "/api/v3/series/lookup?term=tvdb:254376", &lookupResults); err != nil {
 		return 0, fmt.Errorf("lookup series tvdb:254376: %w", err)
 	}
 
@@ -228,7 +229,7 @@ func configureSonarr(qbtPort int) (int, error) {
 		Title string `json:"title"`
 	}
 
-	if err := c.post("/api/v3/series", lookupSeries, &addedSeries); err != nil {
+	if err := c.post(ctx, "/api/v3/series", lookupSeries, &addedSeries); err != nil {
 		return 0, fmt.Errorf("add series: %w", err)
 	}
 
@@ -242,7 +243,7 @@ func configureSonarr(qbtPort int) (int, error) {
 }
 
 // fetchSonarrS01E01 returns the Sonarr episode ID for Season 1, Episode 1 of the given series.
-func fetchSonarrS01E01(seriesID int) (int, error) {
+func fetchSonarrS01E01(ctx context.Context, seriesID int) (int, error) {
 	c := newArrClient(sonarrBase, sonarrAPIKey)
 
 	var episodes []struct {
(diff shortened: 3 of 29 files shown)


e2e/stub/qbittorrent/server.go
```diff
@@ -0,0 +1,317 @@
...
+	// FNV-1a 64-bit, zero-padded to 40 hex chars (20 bytes).
+	var h uint64 = 14695981039346656037
+
+	for _, b := range []byte(name) {
+		h ^= uint64(b)
+		h *= 1099511628211
+	}
+
+	return fmt.Sprintf("%040x", h)
+}
```
solidDoWant: For the purpose of e2e tests, can this just be a static value?

e2e/docker-compose.yml
```diff
@@ -0,0 +1,109 @@
+services:
+  postgres:
+    image: postgres:18
+    command: postgres -c 'max_connections=1000'
```
solidDoWant: Is this line necessary? Why does `max_connections=1000` need to be set?

e2e/docker-compose.yml
```diff
@@ -0,0 +1,109 @@
...
+      - e2e_hatchet_config:/hatchet/config
+    depends_on:
+      setup-config:
+        condition: service_completed_successfully
+      migration:
+        condition: service_completed_successfully
+
+  radarr:
+    image: ghcr.io/home-operations/radarr:6.1.2
+    user: "0:0"
```
solidDoWant: Why is this necessary?

e2e/docker-compose.yml
```diff
@@ -0,0 +1,109 @@
...
+    volumes:
+      - /tmp/e2e-media-processor/config/radarr:/config
+      - /tmp/e2e-media-processor/processed-output/radarr:/downloads
+      - /tmp/e2e-media-processor/processed-output/radarr-library:/movies
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
+
+  sonarr:
+    image: ghcr.io/home-operations/sonarr:4.0.17
+    user: "0:0"
```
solidDoWant: Why is this necessary?

pkg/ffmpeg/stream_video.go
```diff
@@ -362,6 +362,7 @@ func (vss *videoStreamState) setupCropFilter(inStream *astiav.Stream, hwAccel HW
 			devCtx, err := astiav.CreateHardwareDeviceContext(profile.deviceType, vss.hardwareDevicePath, nil, 0)
 			if err != nil {
 				slog.Debug("ffmpeg: VAAPI device context unavailable for crop filter, using SW-only crop", "error", err)
+
```
solidDoWant, line 365: Why was this added?

e2e/e2e_test.go
solidDoWant, line 1: This file is enormous. Would it make sense to split it up into several smaller files?

e2e/e2e_test.go
```diff
@@ -0,0 +1,1089 @@
...
+	sonarrEpisodeID int    // Sonarr episode ID for S01E01 of the test series
+	fixturePath     string // path to BBB mp4 fixture
+	binDir          string // temp dir holding built binaries
+	watcherCmd      *exec.Cmd
+	workerCmd       *exec.Cmd
+)
+
+func TestMain(m *testing.M) {
+	os.Exit(run(m))
+}
```
solidDoWant, line 56: Why is this necessary?

e2e/e2e_test.go
```diff
@@ -0,0 +1,1089 @@
...
+
+func run(m *testing.M) int {
```
solidDoWant: Why return an int instead of an error? Wouldn't most return cases be much simpler if this just returned an error?

e2e/e2e_test.go
```diff
@@ -0,0 +1,1089 @@
...
+		log.Printf("e2e: compose up: %v", err)
+		composeDown()
+
+		return 1
+	}
+
+	defer composeDown()
+
+	// 5. Wait for Radarr, Sonarr, and Hatchet to be healthy.
+	// 10-minute timeout to accommodate image pulls on first run.
```
solidDoWant: What do you think about a separate `docker compose pull` step with a semi-long timeout, and a shorter one for service start?

e2e/e2e_test.go
```diff
@@ -0,0 +1,1089 @@
...
+	base := filepath.Base(dir)
+
+	cmd := exec.Command("docker", "run", "--rm",
+		"-v", parent+":/target-parent",
+		"alpine", "rm", "-rf", "/target-parent/"+base,
+	)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	_ = cmd.Run()
+}
```
solidDoWant: I really don't like this, or the need to run any part of this test as root. Where does the "run as root" requirement come from?

e2e/e2e_test.go
```diff
@@ -0,0 +1,1089 @@
...
+	if out, err := cmd.CombinedOutput(); err != nil {
+		_ = os.Remove(mkvPath)
+
+		return "", fmt.Errorf("create MKV fixture: %w\n%s", err, out)
+	}
+
+	log.Printf("e2e: MKV fixture cached at %s", mkvPath)
+
+	return mkvPath, nil
+}
```
solidDoWant: I kinda understand why this is needed, but I don't really like it. I would prefer if radarr used the full downloaded file as-is.

I understand that this doesn't work well for sonarr because these chosen show is much longer than the fixture's runtime. What do you think about picking another show with a shorter runtime? If you implement this, you should delegate picking a new show/episode to a subagent. Important: any chosen show MUST be in the public domain.

e2e/e2e_test.go
```diff
@@ -0,0 +1,1089 @@
...
+		fn   func() error
+	}
+
+	services := []svc{
+		{"radarr", func() error { return checkHTTP(radarrBase + "/ping") }},
+		{"sonarr", func() error { return checkHTTP(sonarrBase + "/ping") }},
+		{"hatchet-grpc", func() error { return checkTCP("localhost:7079") }},
+	}
+
+	for _, s := range services {
```
solidDoWant: Throughout this whole PR, don't use single letter loop instance var names. Loop instance var names should be the singular version of the loop var. For example, instead of `s`, `service` should be used for a loop far of `services`.

Different instances of the agent keep making the same mistake. A rule should be added to CLAUDE.md or similar to prevent this from happening again.

e2e/e2e_test.go
```diff
@@ -0,0 +1,1089 @@
...
+		_ = os.RemoveAll(binDir)
+	}
+}
+
+func buildBinary(moduleRoot, pkg, out string) ([]byte, error) {
+	cmd := exec.Command("go", "build", "-o", out, pkg)
+	cmd.Dir = moduleRoot
+
+	return cmd.CombinedOutput()
+}
```
solidDoWant: I think this should probably use the makefile target. What do you think?

e2e/e2e_test.go
```diff
@@ -0,0 +1,1089 @@
...
+
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(interval):
+		}
+	}
+}
+
+// ---- Tests --------------------------------------------------------------
```
solidDoWant: Something to keep in mind: in a future PR, additional tests will be added to verify some edge cases, like black-bar cropping. When making architectural decisions, you don't need to design around this, but all else being equal try not to take paths that would make adding more tests difficult.

e2e/e2e_test.go
```diff
@@ -0,0 +1,1089 @@
...
+	require.NoError(t, err, "Radarr did not import the movie within the timeout")
+
+	// Source .mp4 must be deleted by the worker's cleanup step.
+	sourceMp4 := filepath.Join(downloadsDir, "radarr", releaseTitle+".mp4")
+	_, statErr := os.Stat(sourceMp4)
+
+	assert.True(t, os.IsNotExist(statErr), "source .mp4 should have been deleted after import")
+
+	// .mkv must exist somewhere under the Radarr library directory.
+	foundMKV := findMKV(t, filepath.Join(processedDir, "radarr-library"))
```
solidDoWant: This and the sonarr tests should verify expected values on the found output file (like runtime, resolution, codec, container, etc)

e2e/e2e_test.go
```diff
@@ -0,0 +1,1089 @@
...
+	var found bool
+
+	err := filepath.WalkDir(dir, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if !d.IsDir() && strings.HasSuffix(strings.ToLower(path), ".mkv") {
+			found = true
+		}
```
solidDoWant: This should probably return `fs.SkipAll` here

e2e/stub/qbittorrent/server.go
solidDoWant: Done — replaced `syntheticHash` (FNV-1a) with `extractBTIH`, which reads the `xt=urn:btih:` parameter directly from the magnet URI. The tests already construct magnets with explicit btih values (`%040x` of 1 and 2), so no hashing is needed at all.

e2e/docker-compose.yml
solidDoWant: Added a comment explaining the reason. Hatchet's workflow engine opens many concurrent database connections and the default PostgreSQL limit of 100 causes connection errors under load.

e2e/docker-compose.yml
solidDoWant: The original motivation was that Radarr and Sonarr containers (running as root) would create files in the volume-mounted directories owned by root, which the test process (running as the coder user) then couldn't delete. Changed to `user: "1000:1000"` (the coder user's UID/GID) so all container-created files are owned by the current user. `forceRemoveDir` has been removed as a result.

e2e/docker-compose.yml
solidDoWant: Same as the Radarr comment above — changed to `user: "1000:1000"` for the same reason.

pkg/ffmpeg/stream_video.go
solidDoWant: Removed — it was accidentally added while editing the surrounding context. The blank line had no semantic meaning.

e2e/e2e_test.go
solidDoWant: Done — split into 6 focused files: `e2e_test.go` (consts/vars/TestMain/run), `setup_test.go` (directories, Docker Compose, services, processes, polling), `fixture_test.go` (BBB download/cache), `configure_test.go` (Radarr/Sonarr API client and setup), `helpers_test.go` (findMKV, probeOutputFile), `radarr_test.go`, and `sonarr_test.go`.

e2e/e2e_test.go
solidDoWant: Changed to the error-returning pattern. `TestMain` now calls `os.Exit(1)` only when `run()` returns an error; when tests pass, it returns normally. The `os.Exit` wrapper is no longer needed as a top-level call.

e2e/e2e_test.go
solidDoWant: Changed to `error`. All setup failure paths are now `return fmt.Errorf(...)`, and the final line is `if code := m.Run(); code != 0 { return fmt.Errorf(...) }`. This is much cleaner — the log+return pattern replaces the log.Printf+return 1 pattern throughout.

e2e/e2e_test.go
solidDoWant: Done — added `composePull(ctx)` as a separate step with a 20-minute timeout before `composeUp()`. The service-health wait (`waitForServices`) now uses a shorter 5-minute timeout since images are already present by the time the containers start.

e2e/e2e_test.go
solidDoWant: The requirement came from the containers running as root (`user: "0:0"`), which made their volume-written files owned by root — `os.RemoveAll` in the test process (coder user) would fail. Fixed at the source: changed both Radarr and Sonarr to `user: "1000:1000"` so all container-written files are owned by the current user. `forceRemoveDir` and its Docker-based rm-as-root workaround are now removed entirely.

e2e/e2e_test.go
solidDoWant: Done. The qbt stub now uses the raw BBB MP4 for both tests. For Sonarr, switched from The Lone Ranger (TVDB 72059, 26-min episodes) to Colonel Bleep (TVDB 254376, 6-min episodes, public domain — copyright lapsed without renewal in 1985). BBB at ~9:56 is 166% of the 6-minute episode runtime, clearing Sonarr's 50% sample check with margin. The `ensureShortMKVFixture` step has been removed entirely. Per-test timeout increased to 20 minutes to accommodate a full H.265 software encode of the ~9:56 BBB fixture on machines without hardware acceleration.

e2e/e2e_test.go
solidDoWant: Fixed throughout the PR — `s` → `service`, `f` → `file`/`entry`, `h` → `hash`, `k`/`v` → `name`/`cat`, `ep` → `episode`. Also added a rule to `CLAUDE.md` to prevent recurrence: "Loop iteration variable names must be the singular form of the collection name, not single-letter abbreviations."

e2e/e2e_test.go
solidDoWant: Done — `startProcesses` now calls `make build` in the module root instead of `go build` directly. The binaries land in `bin/watcher` and `bin/worker` alongside normal development builds, so no temp dir is needed.

e2e/e2e_test.go
solidDoWant: Noted. The test bodies are now in separate `radarr_test.go` and `sonarr_test.go` files, each with its own polling context and assertions, so adding new test functions (e.g. for crop detection) to either file should be straightforward without touching shared setup code.

e2e/e2e_test.go
solidDoWant: Added. Both tests now run `ffprobe` on the imported .mkv file and assert container (`matroska`), video codec (`hevc`), and that duration is greater than 300 seconds. The helper lives in `helpers_test.go` as `probeOutputFile`.

e2e/e2e_test.go
solidDoWant: Done — `findMKV` now returns `fs.SkipAll` immediately after setting `found` and also returns the path as a `string` instead of `bool`, so callers can pass it directly to `probeOutputFile`.

e2e/docker-compose.yml
solidDoWant: > under load

And will the e2e test produce this much load?

e2e/e2e_test.go
solidDoWant: Short names ("cat" for "category") shouldn't be used unless the loop var is plural of said short name ("cats").

e2e/stub/qbittorrent/server.go
```diff
@@ -0,0 +1,323 @@
...
+
+	s.mu.Unlock()
+
+	log.Printf("qbt stub: torrents/info (query=%s) → %d torrents from %s", r.URL.RawQuery, len(list), r.RemoteAddr)
+
+	w.Header().Set("Content-Type", "application/json")
+	_ = json.NewEncoder(w).Encode(list)
+}
+
+func (s *Server) handleTorrentsAdd(w http.ResponseWriter, r *http.Request) {
```
solidDoWant, line 198: Here and elsewhere in the stub, under the assumption that only the e2e tests will be making API calls to the qbt stub server, can anything be simplified? As an example, does a list of real categories created by `arr` actually need to be tracked, or can these operations just return a simple 200 response?

e2e/setup_test.go
```diff
@@ -0,0 +1,291 @@
...
+		filepath.Join(configDir, "radarr"),
+		filepath.Join(configDir, "sonarr"),
+	} {
+		if err := os.MkdirAll(dir, 0o755); err != nil {
+			return fmt.Errorf("mkdir %s: %w", dir, err)
+		}
+	}
+
+	return nil
+}
```
solidDoWant, line 44: Is this fun still necessary?

e2e/docker-compose.yml
solidDoWant: Good catch on the wording. The e2e test workload itself is minimal (1–2 workflow runs), but Hatchet uses PostgreSQL as its message queue (`SERVER_MSGQUEUE_KIND: postgres`), which means its engine maintains many concurrent DB connections for internal pollers and LISTEN/NOTIFY listeners regardless of how many workflows are active. The connection overhead is a property of Hatchet's initialization, not the test traffic volume. Updated the comment to say this, and also lowered the value from 1000 to 200 — that should be a realistic ceiling for a single engine instance.

e2e/e2e_test.go
solidDoWant: Agreed. After the categories simplification (see the other thread), the loop over `s.categories` no longer exists, so `cat` is gone entirely. Also tightened the CLAUDE.md rule to explicitly prohibit shortened names like `cat` for `category` — it now says: use the full singular form (`category` when ranging over `categories`), and shortened names are only acceptable when the collection itself uses that abbreviation (e.g. `cat` is fine when ranging over `cats`).

e2e/stub/qbittorrent/server.go
solidDoWant: Done. `handleTorrentsCategories` now returns a static `{}` and `handleTorrentsCreateCategory` just logs and returns 200 — no storage. The `Category` type and `categories` map on `Server` are both removed. The category filtering in `handleTorrentsInfo` is unaffected because it filters on the `Category` field stored directly on each `Torrent` at add-time, not on the tracked categories list.

e2e/setup_test.go
solidDoWant: Still necessary, but simplified. The `os.RemoveAll` inside it was redundant (step 0 `composeDown()` already removes `baseDir`), so that's now removed. What remains — the `os.MkdirAll` loop — is still required: Docker creates missing bind-mount directories automatically, but it creates them owned by root, which means the Radarr/Sonarr containers running as `user: "1000:1000"` can't write to them. Pre-creating the directories as the coder user ensures correct ownership before compose starts.

e2e/docker-compose.yml
solidDoWant: What is the default value? Does this need to be explicitly set at all?

e2e/stub/qbittorrent/server.go
solidDoWant: Reverting the simplification — the static `{}` broke the e2e test. Radarr calls `GET /api/v2/torrents/categories` after `POST /api/v2/torrents/createCategory` and rejects the download client with "Configuration of category failed" if the category is not in the response. Category tracking is required for the stub to work correctly. Restored the `Category` type and `categories` map, with `category` as the loop variable name.

e2e/docker-compose.yml
```diff
@@ -0,0 +1,117 @@
...
+        condition: service_completed_successfully
+
+  radarr:
+    image: ghcr.io/home-operations/radarr:6.1.2
+    # Run as the host user so volume-mounted directories are owned by the
+    # current user and can be cleaned up without requiring elevated privileges.
+    user: "1000:1000"
+    environment:
+      TZ: UTC
+      Radarr__Auth__ApiKey: e2e-radarr-api-key
```
solidDoWant: Radarr and sonarr wants an API key that is at least 20 chars long per the logs

e2e/e2e_test.go
```diff
@@ -0,0 +1,154 @@
+//go:build e2e
```
solidDoWant, line 1: Is a build tag really necessary for this package?

e2e/e2e_test.go
```diff
@@ -0,0 +1,154 @@
...
+// watcher and worker subprocesses, and verifies the full happy-path flow.
+//
+// Run with: make test-e2e
+// Prerequisites: Docker, internet access (first run downloads the BBB fixture).
+package e2e_test
+
+import (
+	"context"
+	"fmt"
+	"log"
```
solidDoWant: Use slog everywhere to match the application itself.  This should make  logs easier to read

e2e/setup_test.go
```diff
@@ -0,0 +1,287 @@
...
+func composeArgs(subcmd ...string) []string {
+	args := []string{"compose", "-p", "e2e-media-processor", "-f", "docker-compose.yml"}
+
+	return append(args, subcmd...)
+}
+
+// composePull pulls all Docker images declared in docker-compose.yml.
+// It uses ctx to enforce a timeout for potentially large image downloads.
+func composePull(ctx context.Context) error {
+	cmd := exec.CommandContext(ctx, "docker", composeArgs("pull")...)
```
solidDoWant, line 68: exec.Command* should log to slog instead of stdout/stderr, to make messages easier to read

e2e/docker-compose.yml
solidDoWant: The default PostgreSQL `max_connections` is 100. It does need to be explicitly set — Hatchet's postgres message queue backend opens more than 100 concurrent connections during initialization (for internal pollers, LISTEN/NOTIFY listeners, etc.), regardless of how many workflows are running. The project's dev `docker-compose.yml` uses 1000 for the same reason; 200 is a realistic ceiling for a single engine instance in the e2e suite.

e2e/docker-compose.yml
solidDoWant: Fixed — extended to `e2e-radarr-api-key-e2e` (22 chars) and `e2e-sonarr-api-key-e2e` (22 chars) in both `docker-compose.yml` and `e2e_test.go`.

e2e/e2e_test.go
solidDoWant: Yes — `make test` runs `go test ./...`, which without the build tag would include `./e2e/...` and attempt to start Docker services in any dev or CI environment, failing immediately. The build tag prevents accidental runs and mirrors the `//go:build integration` pattern used for integration tests elsewhere in the project. `make test-e2e` explicitly passes `-tags=e2e` to opt in.

e2e/e2e_test.go
solidDoWant: Done — replaced all `log.Printf`/`log.Println` calls with `slog.Info`/`slog.Error` throughout `e2e_test.go`, `setup_test.go`, `configure_test.go`, and `fixture_test.go`. Messages now use structured key/value pairs (e.g. `slog.Info("watcher started", "pid", ...)`).

e2e/setup_test.go
solidDoWant: Done — added a `slogWriter` (line-buffered `io.Writer`) in `helpers_test.go` that splits output on newlines and emits each line to slog tagged with a `source` attribute. All `cmd.Stdout`/`cmd.Stderr` assignments now use `newSlogWriter(slog.LevelInfo, "source")` and `newSlogWriter(slog.LevelWarn, "source")` for `docker`, `make`, `watcher`, and `worker` commands.

e2e/e2e_test.go
solidDoWant: I'm still seeing a lot of instances of this in the qbt stub

solidDoWant: Fixed — replaced all `log.Printf` calls in `e2e/stub/qbittorrent/server.go` with `slog.Info`/`slog.Warn` using structured key/value pairs (e.g. `slog.Info("torrents/add", "category", category, "releaseName", releaseName, ...)`). The `"log"` import is now gone from that file.

solidDoWant: Shouldn't the qbt logs contain a source field?

Is this missing anywhere else?

solidDoWant: Fixed in two places:

- **qbt stub** (`e2e/stub/qbittorrent/server.go`): added a package-level `var logger = slog.Default().With("source", "qbt")` and switched all `slog.Info`/`slog.Warn` calls to use it. Every stub log line now carries `source=qbt`.

- **Test harness** (`e2e_test.go`, `setup_test.go`, `configure_test.go`, `fixture_test.go`): the direct `slog.Info`/`slog.Error` calls there were also missing `source`. Added a package-level `var log = slog.Default().With("source", "e2e")` in `e2e_test.go` and switched all test-harness calls to use it.

The only slog calls that still go through the top-level `slog` package are in `helpers_test.go` (`slog.Log` inside `slogWriter.Write` for subprocess line routing) — those already carry the per-process `source` attribute passed to `newSlogWriter`.

solidDoWant: Make the qbt log entries "debug"

pkg/logging/logging.go
```diff
@@ -32,3 +34,31 @@ func Setup(level string) {
...
+	default:
+		return "info"
+	}
+}
+
+// NewZerologLogger returns a zerolog.Logger writing JSON to stderr, configured
+// at the level derived from the given LOG_LEVEL string and tagged with the
+// given service name. Use this to configure Hatchet SDK worker loggers so they
+// respect LOG_LEVEL rather than always defaulting to debug.
+func NewZerologLogger(level, service string) zerolog.Logger {
```
solidDoWant: Does LOG_LEVEL need to be set here? Why not configure zerolog with the highest level, and let slog filter it?

cmd/watcher/main.go
```diff
@@ -53,7 +54,9 @@ func run(ctx context.Context, configPath string) error {
 	}
 	defer shutdown()
 
-	client, err := hatchet.NewClient()
+	client, err := hatchet.NewClient(
+		v0Client.WithLogLevel(logging.ZerologLevel(os.Getenv("LOG_LEVEL"))), //nolint:staticcheck // no non-deprecated API for client log level in SDK v0.83
+	)
```
solidDoWant, line 61: Will the client use the slog logger? Is there a way to make it use the slog logger if not?

e2e/e2e_test.go
solidDoWant: Done — changed all `logger.Info` and `logger.Warn` calls in `e2e/stub/qbittorrent/server.go` to `logger.Debug`.

pkg/logging/logging.go
solidDoWant: Good call. `zerologSlogBridge.Write` already calls `slog.Default().Enabled(ctx, lvl)` before forwarding, so zerolog's own level filter is redundant. Removed the `level` parameter from `NewZerologLogger` and the `ZerologLevel` helper entirely; zerolog is now configured at `zerolog.TraceLevel` so every entry passes through to the bridge, which lets slog do the filtering.

cmd/watcher/main.go
solidDoWant: Yes — `NewZerologLogger` returns a `zerolog.Logger` whose writer is `zerologSlogBridge`. That bridge parses each zerolog JSON entry and re-emits it via `slog.Log(...)`, so the Hatchet client's internal logs go through the default slog handler with the same formatting and level filtering as the rest of the application. After the change to address the `LOG_LEVEL` comment above, level filtering is now done entirely by slog; zerolog runs at `TraceLevel` and passes everything through.

cmd/watcher/main.go
```diff
@@ -65,7 +70,10 @@ func run(ctx context.Context, configPath string) error {
 		return fmt.Errorf("create scan workflow: %w", err)
 	}
 
+	watcherLogger := logging.NewZerologLogger("worker")
```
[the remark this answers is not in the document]
solidDoWant: I'm not sure if this is correct in the current context. Need to verify one way or the other before replying.

e2e/docker-compose.yml
```diff
@@ -0,0 +1,117 @@
...
+    command: postgres -c 'max_connections=200'
+    hostname: postgres
+    environment:
+      POSTGRES_USER: hatchet
+      POSTGRES_PASSWORD: hatchet
+      POSTGRES_DB: hatchet
+    ports:
+      - "5436:5432"
+    volumes:
+      - e2e_hatchet_postgres_data:/var/lib/postgresql
```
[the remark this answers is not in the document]
solidDoWant: I think this very recently changed and I'm no longer sure which approach is right.

e2e/stub/qbittorrent/server.go
```diff
@@ -0,0 +1,338 @@
...
+
+func (s *Server) handleTorrentsInfo(w http.ResponseWriter, r *http.Request) {
+	categoryFilter := r.URL.Query().Get("category")
+
+	s.mu.Lock()
+
+	list := make([]*Torrent, 0, len(s.torrents))
+	for _, t := range s.torrents {
+		if categoryFilter == "" || t.Category == categoryFilter {
+			list = append(list, t)
```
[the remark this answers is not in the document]
solidDoWant: Fixed — renamed `t` to `torrent` in the `for _, torrent := range s.torrents` loop.

pkg/ffmpeg/transcode_test.go
```diff
@@ -677,6 +678,122 @@ func TestWithoutCrop_DimensionsUnchanged(t *testing.T) {
...
+
+	var videoStream *ffprobe.StreamInfo
+
+	for i, s := range outputInfo.Streams {
+		if s.CodecType == ffprobe.CodecTypeVideo {
+			videoStream = &outputInfo.Streams[i]
+			break
+		}
+	}
+
```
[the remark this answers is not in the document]
solidDoWant: Fixed — renamed `s` → `stream` in all three loops (`for i, stream := range outputInfo.Streams`, `for i, stream := range inputInfo.Streams`, `for _, stream := range fmtCtx.Streams()`).

pkg/ffmpeg/transcode_test.go
```diff
@@ -677,6 +678,122 @@ func TestWithoutCrop_DimensionsUnchanged(t *testing.T) {
...
+	var packetCount int
+
+	// Collect PTS values to verify they form a valid, monotonic display order.
+	var ptsValues []int64
+
+	for {
+		if err := fmtCtx.ReadFrame(pkt); err != nil {
+			break
+		}
```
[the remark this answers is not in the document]
solidDoWant: Fixed — the loop now checks `errors.Is(err, astiav.ErrEof)` and calls `require.True(t, ...)` to fail the test on any non-EOF error, matching the pattern used in the production `transcode.go:814` and `cropdetect.go:340` read loops.

cmd/watcher/main.go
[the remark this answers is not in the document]
solidDoWant: Verified and fixed. Both `cmd/watcher` and `cmd/worker` previously used `service=worker` for their internal Hatchet SDK workers, so aggregated logs from both binaries would be ambiguous. The watcher binary's internal worker logger is now `NewZerologLogger("watcher")`, so its Hatchet SDK logs carry `source=watcher` instead. The worker binary's logger stays as `"worker"`.

e2e/docker-compose.yml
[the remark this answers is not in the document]
solidDoWant: Changed to `/var/lib/postgresql/data`. The official postgres image sets `ENV PGDATA /var/lib/postgresql/data`, so that is the correct and conventional mount point for the data volume. Mounting at the parent `/var/lib/postgresql` would work (PostgreSQL would create `data/` inside the volume), but mounting directly at `PGDATA` is the standard approach and matches what the copilot suggested.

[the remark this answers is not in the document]
solidDoWant: Fixed — changed to `/var/lib/postgresql/data` to match the official postgres image's `PGDATA` default.

e2e/docker-compose.yml
```diff
@@ -0,0 +1,117 @@
...
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
+
+  sonarr:
+    image: ghcr.io/home-operations/sonarr:4.0.17
+    # Run as the host user so volume-mounted directories are owned by the
+    # current user and can be cleaned up without requiring elevated privileges.
+    user: "1000:1000"
+    environment:
+      TZ: UTC
```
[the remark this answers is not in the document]
solidDoWant: Intentional for this project. The e2e suite is designed to run in the coder dev environment where the developer user is always UID/GID 1000. Using `${UID}:${GID}` would require env var substitution in docker-compose.yml, which the project explicitly avoids (the plan calls for a fully static compose file with no variable substitution). Omitting `user` entirely would cause the containers to run as root, making bind-mount files owned by root and breaking `os.RemoveAll` cleanup — which is exactly the problem this `user` field was added to fix.

e2e/docker-compose.yml
```diff
@@ -0,0 +1,117 @@
...
+        condition: service_completed_successfully
+      migration:
+        condition: service_completed_successfully
+
+  radarr:
+    image: ghcr.io/home-operations/radarr:6.1.2
+    # Run as the host user so volume-mounted directories are owned by the
+    # current user and can be cleaned up without requiring elevated privileges.
+    user: "1000:1000"
+    environment:
```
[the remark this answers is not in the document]
solidDoWant: Same as the Radarr comment above — intentional for the coder dev environment (always UID/GID 1000). A static compose file with no variable substitution was an explicit design decision for this suite.

e2e/docker-compose.yml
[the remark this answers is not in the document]
solidDoWant: in this case it's probably okay

e2e/docker-compose.yml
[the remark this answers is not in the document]
solidDoWant: in this case it's probably okay

e2e/docker-compose.yml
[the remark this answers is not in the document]
solidDoWant: Updated — `docker-compose.yml` now uses `user: "${UID}:${GID}"` for both Radarr and Sonarr. Since `UID`/`GID` are not exported by bash by default, `setup_test.go` injects them via a new `composeEnv()` helper that appends `os.Getuid()`/`os.Getgid()` to the process environment on every compose command.

e2e/docker-compose.yml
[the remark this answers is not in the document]
solidDoWant: Updated — same change as the Radarr thread above. Both services now use `${UID}:${GID}`.

e2e/stub/qbittorrent/server.go
```diff
@@ -0,0 +1,338 @@
...
+
+	if err := os.MkdirAll(destDir, 0o755); err != nil {
+		http.Error(w, fmt.Sprintf("mkdir %s: %v", destDir, err), http.StatusInternalServerError)
+
+		return
+	}
+
+	destPath := filepath.Join(destDir, releaseName+".mp4")
+	tmpPath := destPath + ".tmp"
+
```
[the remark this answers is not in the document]
solidDoWant: not sure that this is relevant in the scope of an e2e test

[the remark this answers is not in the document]
solidDoWant: Agreed with the owner's assessment — no change made. The stub only receives connections from the e2e test harness and the arr containers (both running on a controlled localhost network), so arbitrary attacker-controlled input is not a realistic threat here.

e2e/setup_test.go
```diff
@@ -0,0 +1,305 @@
...
+}
+
+func checkHTTP(url string) error {
+	resp, err := http.Get(url) //nolint:noctx // health poll, no caller context
+	if err != nil {
+		return err
+	}
+
+	_ = resp.Body.Close()
+
```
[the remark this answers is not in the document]
solidDoWant: Fixed — `checkHTTP` now uses `&http.Client{Timeout: 2 * time.Second}` and returns an error for any 4xx/5xx response.

e2e/fixture_test.go
```diff
@@ -0,0 +1,130 @@
...
+	f, err := os.Create(dest)
+	if err != nil {
+		return err
+	}
+
+	defer func() { _ = f.Close() }()
+
+	_, err = io.Copy(f, resp.Body)
+
+	return err
```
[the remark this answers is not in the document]
solidDoWant: Fixed — `downloadFile` now accepts a `context.Context` and uses `http.NewRequestWithContext` so a stalled or slow download is bounded by the caller's deadline. `ensureBBBFixture` also accepts a context; `run()` wraps the call in a 30-minute `context.WithTimeout` so the suite fails fast with a clear error rather than hanging until the global test timeout fires.

pkg/logging/logging.go
```diff
@@ -32,3 +37,67 @@ func Setup(level string) {
...
+	switch strings.ToLower(level) {
+	case "debug":
+		return slog.LevelDebug
+	case "warn", "warning":
+		return slog.LevelWarn
+	case "error", "fatal", "panic":
+		return slog.LevelError
+	default:
+		return slog.LevelInfo
+	}
```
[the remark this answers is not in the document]
solidDoWant: Fixed — added `case "trace"` to `zerologLevelToSlog` mapping it to `slog.LevelDebug`. Trace entries are now correctly forwarded at debug level rather than being silently promoted to info.

e2e/helpers_test.go
```diff
@@ -0,0 +1,134 @@
...
+		}
+
+		w.buf = w.buf[idx+1:]
+	}
+
+	return len(p), nil
+}
+
+// newSlogWriter returns an io.Writer that routes subprocess output to slog.
+// Use LevelInfo for stdout and LevelWarn for stderr.
```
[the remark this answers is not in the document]
solidDoWant: I don't know if this is critical given the use case

e2e/stub/qbittorrent/server.go
```diff
@@ -0,0 +1,338 @@
...
+	magnetStr := r.FormValue("urls")
+	if magnetStr == "" {
+		magnetStr = r.FormValue("magnet")
+	}
+
+	releaseName := extractDN(magnetStr)
+	if releaseName == "" {
+		releaseName = "unknown-release"
+	}
+
```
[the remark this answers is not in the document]
solidDoWant: I don't know if this is critical given the use case

e2e/helpers_test.go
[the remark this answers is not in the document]
solidDoWant: Fixed — added a `Flush()` method to `slogWriter` that emits any buffered partial line. Changed `newSlogWriter` to return `*slogWriter` (still satisfies `io.Writer`) so callers can call `Flush()`. Updated `composePull`, `composeUp`, `composeDown`, and the `make build` step in `startProcesses` to call `Flush()` after each command exits.

pkg/ffmpeg/transcode_test.go
```diff
@@ -677,6 +679,124 @@ func TestWithoutCrop_DimensionsUnchanged(t *testing.T) {
...
+	sorted := make([]int64, len(ptsValues))
+	copy(sorted, ptsValues)
+
+	sort.Slice(sorted, func(i, j int) bool { return sorted[i] < sorted[j] })
+
+	for i := 1; i < len(sorted); i++ {
+		assert.Greater(t, sorted[i], sorted[i-1],
+			"display-order PTS must be strictly increasing: pts[%d]=%d, pts[%d]=%d",
+			i-1, sorted[i-1], i, sorted[i])
+	}
```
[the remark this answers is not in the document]
solidDoWant: Fixed — replaced the sort+monotonicity check with a set-based uniqueness check. Sorting made the strictly-increasing assertion trivially true (always passes after sort), so the only thing it was actually catching was duplicate PTS values. The new check uses a `map[int64]bool` to explicitly assert no two packets share a PTS value, which is the actual invariant being validated. Also removed the now-unused `sort` import.

cmd/worker/main.go
```diff
@@ -9,6 +9,7 @@ import (
 	"strconv"
 	"syscall"
 
+	v0Client "github.com/hatchet-dev/hatchet/pkg/client" //nolint:staticcheck // needed for WithLogLevel; no new-SDK equivalent
```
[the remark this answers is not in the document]
solidDoWant: Fixed — updated comment to `// needed for WithLogger; no new-SDK equivalent`.

cmd/watcher/main.go
```diff
@@ -9,6 +9,7 @@ import (
 	"os/signal"
 	"syscall"
 
+	v0Client "github.com/hatchet-dev/hatchet/pkg/client" //nolint:staticcheck // needed for WithLogLevel; no new-SDK equivalent
```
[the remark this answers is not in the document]
solidDoWant: Fixed — updated comment to `// needed for WithLogger; no new-SDK equivalent`.

e2e/stub/qbittorrent/server.go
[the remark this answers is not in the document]
solidDoWant: Same situation as the earlier `category` thread (`r3084323713`) — the `dn` parameter comes from a magnet URI constructed by the e2e test harness itself (hardcoded strings like `Big.Buck.Bunny.2008.1080p`), not from arbitrary external input. No change made.

e2e/setup_test.go
```diff
@@ -0,0 +1,311 @@
...
+	return cmd.Run()
+}
+
+func composeUp() error {
+	cmd := exec.Command("docker", composeArgs("up", "-d")...)
+	cmd.Env = composeEnv()
+	cmd.Stdout = newSlogWriter(slog.LevelInfo, "docker")
+	cmd.Stderr = newSlogWriter(slog.LevelWarn, "docker")
+
+	return cmd.Run()
```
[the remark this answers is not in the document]
solidDoWant: Fixed — `composeUp` now uses `exec.CommandContext` with a 2-minute timeout (matching the copilot suggestion). If `docker compose up -d` hangs, the context cancels it and the error propagates to the caller.

e2e/radarr_test.go
```diff
@@ -0,0 +1,92 @@
...
+	assert.Greater(t, info.durationSec, 300.0, "output duration should be at least 5 minutes")
+
+	// debugging: copy the output file to /workspace/media-processor for manual inspection after the test completes
+	debugDest := filepath.Join("/workspace/media-processor", "radarr_happy_path_output.mkv")
+	if err := copyFile(mkvPath, debugDest); err != nil {
+		t.Logf("failed to copy output file for debugging: %v", err)
+		return
+	}
+
+	t.Logf("copied output file to %s for debugging", debugDest)
```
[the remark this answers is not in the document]
solidDoWant: I forgot to remove these blocks. Please remove these copy blocks (and the associated copyFile func) in both tests.

e2e/sonarr_test.go
```diff
@@ -0,0 +1,123 @@
...
+	assert.Greater(t, info.durationSec, 300.0, "output duration should be at least 5 minutes")
+
+	// debugging: copy the output file to /workspace/media-processor for manual inspection after the test completes
+	debugDest := filepath.Join("/workspace/media-processor", "sonarr_happy_path_output.mkv")
+	if err := copyFile(mkvPath, debugDest); err != nil {
+		t.Logf("failed to copy output file for debugging: %v", err)
+		return
+	}
+
+	t.Logf("copied output file to %s for debugging", debugDest)
```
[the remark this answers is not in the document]
solidDoWant: I forgot to remove these blocks. Please remove these copy blocks (and the associated copyFile func) in both tests.

pkg/logging/logging.go
```diff
@@ -32,3 +37,67 @@ func Setup(level string) {
...
+// exactly once per complete log entry, so the bridge can parse and forward
+// without buffering.
+type zerologSlogBridge struct{}
+
+func (b *zerologSlogBridge) Write(p []byte) (int, error) {
+	ctx := context.TODO()
+
+	var raw map[string]any
+	if err := json.Unmarshal(bytes.TrimRight(p, "\n"), &raw); err != nil {
+		slog.Warn("hatchet: non-JSON log entry", "raw", strings.TrimRight(string(p), "\n"))
```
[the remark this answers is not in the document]
solidDoWant: Fixed — added `service string` to `zerologSlogBridge`, passed from `NewZerologLogger`, and updated the fallback warning to `"zerolog: non-JSON log entry"` with `"service", b.service` so the source is identifiable in mixed logs.

pkg/ffmpeg/transcode_test.go
```diff
@@ -677,6 +678,121 @@ func TestWithoutCrop_DimensionsUnchanged(t *testing.T) {
 	assert.Equal(t, 220, videoStream.HeightPixels, "height should be unchanged without crop")
 }
 
+// TestTranscode_H265_VideoTimestampsValid verifies that transcoding to H.265
+// produces output with correct video timestamps: every video packet must have a
+// valid PTS (not AV_NOPTS_VALUE), display-order PTS values must increase
+// monotonically, and the output frame rate must match the input.
```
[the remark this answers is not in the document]
solidDoWant: Fixed — updated the docstring to say "packet PTS values must not be duplicated" instead of "display-order PTS values must increase monotonically", matching the set-based uniqueness check that replaced the sort+monotonicity assertion.

e2e/sonarr_test.go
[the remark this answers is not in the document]
solidDoWant: Removed — the debug copy block and `copyFile` helper are gone from both `radarr_test.go` and `sonarr_test.go`.

e2e/radarr_test.go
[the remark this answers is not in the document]
solidDoWant: Removed — same as the Sonarr thread above.

e2e/configure_test.go
```diff
@@ -0,0 +1,291 @@
...
+	apiKey     string
+	httpClient *http.Client
+}
+
+func newArrClient(base, apiKey string) *arrClient {
+	return &arrClient{base: base, apiKey: apiKey, httpClient: &http.Client{Timeout: 30 * time.Second}}
+}
+
+func (c *arrClient) get(path string, out any) error {
+	req, err := http.NewRequest(http.MethodGet, c.base+path, nil)
```
solidDoWant: `get` and `post` use `http.NewRequest` without a context. When these methods are called from test functions (e.g. inside `pollUntil` closures in `radarr_test.go:56` and `sonarr_test.go:64`), the requests can't be cancelled by the test's context timeout. The 30-second client timeout prevents indefinite hangs, but the polling context's cancellation won't propagate to in-flight HTTP requests.

Consider adding a `ctx context.Context` parameter to both methods and using `http.NewRequestWithContext`. For the `TestMain` call sites (which have no test context), `context.Background()` is fine. For the test-function call sites, callers can pass `t.Context()` or the polling deadline context.

e2e/stub/qbittorrent/server.go
```diff
@@ -0,0 +1,338 @@
...
+	for _, hash := range strings.Split(r.FormValue("hashes"), "|") {
+		hash = strings.TrimSpace(hash)
+		if hash == "" {
+			continue
+		}
+
+		s.mu.Lock()
+		delete(s.torrents, hash)
+		s.mu.Unlock()
```
solidDoWant: Nit: the mutex is acquired and released on every iteration, which means a concurrent `handleTorrentsInfo` could observe a partially-deleted batch. In practice the e2e tests delete one torrent at a time so this is harmless, but holding the lock for the entire loop would be simpler and correct:

```go
s.mu.Lock()
for _, hash := range strings.Split(r.FormValue("hashes"), "|") {
    hash = strings.TrimSpace(hash)
    if hash == "" {
        continue
    }
    delete(s.torrents, hash)
}
s.mu.Unlock()
```

e2e/radarr_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+		return nil
+	})
+	require.NoError(t, err, "Radarr did not import the movie within the timeout")
+
+	// Source .mp4 must be deleted by the worker's cleanup step.
+	sourceMp4 := filepath.Join(downloadsDir, "radarr", releaseTitle+".mp4")
+	_, statErr := os.Stat(sourceMp4)
+
+	assert.True(t, os.IsNotExist(statErr), "source .mp4 should have been deleted after import")
```
solidDoWant: Nit: `os.IsNotExist` is the pre-Go 1.13 form. The modern idiomatic form is `errors.Is(statErr, os.ErrNotExist)`, which handles wrapped errors correctly. Same applies to the identical check in `sonarr_test.go:80`.

e2e/setup_test.go
```diff
@@ -0,0 +1,340 @@
...
+func pollUntil(ctx context.Context, interval time.Duration, cond func() error) error {
+	for {
+		if err := cond(); err == nil {
+			return nil
+		}
+
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(interval):
```
solidDoWant: Nit: `time.After` returns a channel backed by a timer that can't be stopped. On each successful poll iteration (and the final one before the `ctx.Done()` check), the timer goroutine leaks until the interval elapses. For 10-second intervals in a minutes-long test this is harmless, but using `time.NewTimer` with an explicit `Stop()` is the cleaner pattern:

```go
timer := time.NewTimer(interval)
select {
case <-ctx.Done():
    timer.Stop()
    return ctx.Err()
case <-timer.C:
}
```

pkg/ffmpeg/transcode_test.go
solidDoWant, line 1: The two new crop tests (`TestWithCrop_NarrowsOutputDimensions` and `TestWithoutCrop_DimensionsUnchanged`) use `context.Background()` in four places (lines 632, 635, 661, 664). Per the CLAUDE.md testing rules, these should use `t.Context()` instead:

```go
Run(t.Context())
// ...
info, err := ffprobe.Probe(t.Context(), output)
```

solidDoWant (commented):
Solid PR -- the e2e test suite is well-structured and the separation into focused files (`fixture_test.go`, `configure_test.go`, `helpers_test.go`, etc.) makes each piece easy to follow. The qBittorrent stub is a nice approach for avoiding a heavy external dependency. The bug fixes in `artwork.go` (extension fallback from URL), `radarr/client.go` and `sonarr/client.go` (passing `title` instead of raw path to the parse endpoint), and the frame rate propagation in `transcode.go` all look correct and well-motivated by the e2e failures.

A few things I noticed:

**Context propagation in `arrClient`** (the most meaningful one): `get`/`post` methods don't accept a `context.Context`, so HTTP requests made during test polling can't be cancelled by the test's deadline context. The 30s client timeout is a backstop but doesn't integrate with the test lifecycle.

**Minor items**: `context.Background()` instead of `t.Context()` in the new crop tests, `os.IsNotExist` vs the modern `errors.Is` form, lock-per-iteration in the qbt stub delete handler, and a `time.After` timer leak in `pollUntil`. All detailed in inline comments.

The prior review rounds covered a lot of ground and the responses look thorough. Nice work getting both tests passing in under a minute each.

e2e/radarr_test.go
solidDoWant: Or maybe use `assert.ErrorIs`

e2e/configure_test.go
solidDoWant: Done. Added `ctx context.Context` as the first parameter to both `get` and `post`, switching to `http.NewRequestWithContext`. Propagated `ctx` through `configureRadarr`, `configureSonarr`, and `fetchSonarrS01E01` — their `TestMain` call sites pass `context.Background()`. The `pollUntil` closures in `radarr_test.go` and `sonarr_test.go` now pass the polling deadline `ctx`; the initial release-push calls pass `t.Context()`.

e2e/stub/qbittorrent/server.go
solidDoWant: Done. Moved `s.mu.Lock()` / `s.mu.Unlock()` outside the loop so the entire batch deletion is atomic.

e2e/radarr_test.go
solidDoWant: Used `assert.ErrorIs` — cleaner and handles wrapped errors. Applied to both `radarr_test.go` and `sonarr_test.go`.

e2e/setup_test.go
solidDoWant: Done. Replaced `time.After` with `time.NewTimer` + `timer.Stop()` on the context-cancelled path.

pkg/ffmpeg/transcode_test.go
solidDoWant: Done. All four `context.Background()` calls in `TestWithCrop_NarrowsOutputDimensions` and `TestWithoutCrop_DimensionsUnchanged` replaced with `t.Context()`.

pkg/logging/logging.go
```diff
@@ -32,3 +37,69 @@ func Setup(level string) {
...
+type zerologSlogBridge struct {
+	service string
+}
+
+func (b *zerologSlogBridge) Write(p []byte) (int, error) {
+	ctx := context.TODO()
+
+	var raw map[string]any
+	if err := json.Unmarshal(bytes.TrimRight(p, "\n"), &raw); err != nil {
+		slog.Warn("zerolog: non-JSON log entry", "service", b.service, "raw", strings.TrimRight(string(p), "\n"))
```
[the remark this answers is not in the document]
solidDoWant: Done. Added `err` and `raw_len` as structured attributes on the warning log.

e2e/setup_test.go
```diff
@@ -0,0 +1,342 @@
...
+		}
+
+		timer := time.NewTimer(interval)
+		select {
+		case <-ctx.Done():
+			timer.Stop()
+			return ctx.Err()
+		case <-timer.C:
+		}
+	}
```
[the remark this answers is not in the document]
solidDoWant: Done. `pollUntil` now tracks `lastErr` and returns `fmt.Errorf("%w; last condition error: %w", ctx.Err(), lastErr)` on deadline/cancellation, so both `ctx.Err()` and the last cond error are visible to callers.

e2e/radarr_test.go
```diff
@@ -0,0 +1,83 @@
...
+		t.Logf("Radarr push response: %s", pushResp)
+	}
+
+	// Poll until Radarr has imported the movie (hasFile=true).
+	// The timeout is set generously to accommodate a full H.265 software encode
+	// of the BBB fixture on machines without hardware acceleration.
+	ctx, cancel := context.WithTimeout(t.Context(), 60*time.Minute)
+	defer cancel()
+
+	err := pollUntil(ctx, 10*time.Second, func() error {
```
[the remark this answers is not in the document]
solidDoWant: Done. Now marshals `pushResp` to JSON before logging; falls back to `%+v` with the marshal error if that fails.

e2e/sonarr_test.go
```diff
@@ -0,0 +1,91 @@
...
+		"publishDate":        time.Now().UTC().Format(time.RFC3339),
+		"indexer":            "e2e-test",
+		"size":               700_000_000,
+		"mappedSeriesId":     sonarrSeriesID,
+		"mappedSeasonNumber": 1,
+		"mappedEpisodeIds":   []int{sonarrEpisodeID},
+	}, &sonarrPushResp), "push release to Sonarr")
+
+	if len(sonarrPushResp) > 0 {
+		t.Logf("Sonarr push response: %s", sonarrPushResp)
```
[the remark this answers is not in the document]
solidDoWant: Done. Same fix applied — marshals `sonarrPushResp` to JSON before logging, matching the suggested pattern.

e2e/helpers_test.go
```diff
@@ -0,0 +1,146 @@
...
+		"-v", "quiet",
+		"-print_format", "json",
+		"-show_format",
+		"-show_streams",
+		path,
+	)
+
+	out, err := cmd.Output()
+	require.NoError(t, err, "ffprobe failed on %s", path)
+
```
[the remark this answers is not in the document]
solidDoWant: Done. `probeOutputFile` now uses `exec.CommandContext(t.Context(), "ffprobe", ...)` so the process is killed when the test context is cancelled.

PR #102, opened 2026-04-20
feat(health): add HTTP health and readiness probes to watcher and worker

### LINKED ISSUE #101: feat(health): add HTTP health and readiness probes to watcher and worker
## Summary

The `cmd/watcher` and `cmd/worker` binaries expose no HTTP health endpoints, making it impossible for orchestration systems (Kubernetes, Nomad, etc.) to verify that the processes have initialized and are operating correctly. This issue adds `/healthz` (liveness) and `/readyz` (readiness) HTTP endpoints to both binaries via a new `HEALTH_ADDR` environment variable.

## Acceptance Criteria

- [x] **Given** `HEALTH_ADDR` is set to a TCP address, **when** either binary starts, **then** an HTTP server is listening on that address and responds to requests on `/healthz` and `/readyz`.
- [x] **Given** the health server is running, **when** `GET /healthz` is requested at any point after the server starts, **then** the response is `200 OK`.
- [x] **Given** the binary has not yet finished initialization (e.g., still connecting to Hatchet), **when** `GET /readyz` is requested, **then** the response is `503 Service Unavailable`.
- [x] **Given** the binary has successfully completed initialization and the Hatchet worker has started, **when** `GET /readyz` is requested, **then** the response is `200 OK`.
- [x] **Given** `HEALTH_ADDR` is unset or empty, **when** either binary starts, **then** no health HTTP server is started and behavior is unchanged.
- [x] **Given** the new `HEALTH_ADDR` environment variable, **when** `docs/configuration.md` is consulted, **then** the variable is documented with its type, default, required status, and description, consistent with existing entries in that file and in `SPEC.md`.

## Technical Context

- Depends on: none
- Both binaries call `worker.StartBlocking(ctx)`, which blocks until context cancellation or error. The readiness state must be set **before** calling `StartBlocking`, immediately after all initialization succeeds (config loaded, env vars validated, Hatchet client and worker object created).
- `HEALTH_ADDR` should be documented in the "Shared (`cmd/watcher` and `cmd/worker`)" section of the environment variable table in `SPEC.md` and `docs/configuration.md`, consistent with `METRICS_ADDR`.
- The health server must be independent of the existing Prometheus metrics server (`METRICS_ADDR`) so health probes work in deployments that do not configure metrics.
- A new `pkg/health` package is the recommended home. It should provide a type that starts the HTTP server and exposes a method (e.g., `SetReady()`) to atomically flip the readiness state.
- `/healthz` is the conventional liveness path (returns 200 as long as the HTTP server is responding). `/readyz` is the conventional readiness path.
- Readiness is startup-time only: it flips once to ready after successful initialization and does not revert based on runtime connectivity state.

## Non-Goals

- Kubernetes manifests or Helm charts (the `deploy/k8s/` directory is intentionally empty for now).
- Ongoing runtime health-checking of Hatchet, Sonarr, or Radarr connectivity after startup.
- Merging the health endpoints into the existing Prometheus metrics HTTP server.

solidDoWant:
Fixes #101

### DIFF
--- a/cmd/watcher/main.go
+++ b/cmd/watcher/main.go
@@ -12,6 +12,7 @@ import (
 	v0Client "github.com/hatchet-dev/hatchet/pkg/client" //nolint:staticcheck // needed for WithLogger; no new-SDK equivalent
 	hatchet "github.com/hatchet-dev/hatchet/sdks/go"
 
+	"github.com/solidDoWant/media-processor/pkg/health"
 	"github.com/solidDoWant/media-processor/pkg/logging"
 	"github.com/solidDoWant/media-processor/pkg/metrics"
 )
@@ -32,6 +33,11 @@ func main() {
 
 func run(ctx context.Context, configPath string) error {
 	logging.Setup(os.Getenv("LOG_LEVEL"))
+
+	healthServer, err := health.NewFromEnv()
+	if err != nil {
+		return fmt.Errorf("init health server: %w", err)
+	}
 
 	cfg, err := loadConfig(configPath)
 	if err != nil {
@@ -81,6 +87,10 @@ func run(ctx context.Context, configPath string) error {
 	}
 
 	slog.Info("starting directory scan worker", "schedule", cfg.CronSchedule)
+
+	if healthServer != nil {
+		healthServer.SetReady()
+	}
 
 	if err := worker.StartBlocking(ctx); err != nil {
 		return fmt.Errorf("watcher stopped: %w", err)
--- a/cmd/watcher/main.go
+++ b/cmd/watcher/main.go
@@ -34,9 +34,15 @@ func main() {
 func run(ctx context.Context, configPath string) error {
 	logging.Setup(os.Getenv("LOG_LEVEL"))
 
-	healthServer, err := health.NewFromEnv()
-	if err != nil {
-		return fmt.Errorf("init health server: %w", err)
+	var healthServer *health.Server
+
+	if addr := os.Getenv("HEALTH_ADDR"); addr != "" {
+		var err error
+
+		healthServer, err = health.New(ctx, addr)
+		if err != nil {
+			return fmt.Errorf("init health server: %w", err)
+		}
 	}
 
 	cfg, err := loadConfig(configPath)
--- a/pkg/health/health.go
+++ b/pkg/health/health.go
@@ -1,0 +1,62 @@
+package health
+
+import (
+	"errors"
+	"fmt"
+	"net"
+	"net/http"
+	"os"
+	"sync/atomic"
+)
+
+// Server serves /healthz (always 200) and /readyz (503 until SetReady is called).
+type Server struct {
+	ready atomic.Bool
+}
+
+// New starts an HTTP health server on addr. Returns an error if the listener cannot be opened.
+func New(addr string) (*Server, error) {
+	s := &Server{}
+
+	listener, err := net.Listen("tcp", addr)
+	if err != nil {
+		return nil, fmt.Errorf("listen on health addr %s: %w", addr, err)
+	}
+
+	mux := http.NewServeMux()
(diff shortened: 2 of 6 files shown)


cmd/watcher/main.go
```diff
@@ -33,6 +34,11 @@ func main() {
 func run(ctx context.Context, configPath string) error {
 	logging.Setup(os.Getenv("LOG_LEVEL"))
 
+	healthServer, err := health.NewFromEnv()
```
solidDoWant: `NewFromEnv` is inconsistent with the theme of pulling all env vars in run and propagating downward

pkg/health/health.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	mux := http.NewServeMux()
+	mux.HandleFunc("/healthz", func(w http.ResponseWriter, _ *http.Request) {
+		w.WriteHeader(http.StatusOK)
+	})
+	mux.HandleFunc("/readyz", func(w http.ResponseWriter, _ *http.Request) {
+		if s.ready.Load() {
+			w.WriteHeader(http.StatusOK)
+		} else {
```
solidDoWant: nit: replace the else clause with an early return

pkg/health/health.go
```diff
@@ -0,0 +1,62 @@
...
+	"errors"
+	"fmt"
+	"net"
+	"net/http"
+	"os"
+	"sync/atomic"
+)
+
+// Server serves /healthz (always 200) and /readyz (503 until SetReady is called).
+type Server struct {
```
solidDoWant, line 15: Should the server support context cancellation?

cmd/watcher/main.go
solidDoWant: Fixed — removed `NewFromEnv()` from the package entirely. Both binaries now read `HEALTH_ADDR` directly in `run()` and call `health.New(ctx, addr)` conditionally, consistent with how other env vars are handled.

pkg/health/health.go
solidDoWant: Done — replaced with an early return on the not-ready path.

pkg/health/health.go
solidDoWant: Yes, added. `New()` now takes a `context.Context` as its first parameter. A background goroutine calls `srv.Shutdown(context.Background())` when the context is cancelled, so the server stops cleanly on SIGINT/SIGTERM alongside the rest of the process.

pkg/health/health.go
```diff
@@ -0,0 +1,61 @@
...
+		if err := srv.Serve(listener); err != nil && !errors.Is(err, http.ErrServerClosed) {
+			fmt.Fprintf(os.Stderr, "health HTTP server error: %v\n", err)
+		}
+	}()
+
+	go func() {
+		<-ctx.Done()
+
+		_ = srv.Shutdown(context.Background()) //nolint:contextcheck // intentional: process is exiting
+	}()
```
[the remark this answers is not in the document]
solidDoWant: Fixed — the shutdown goroutine now uses a 5-second timeout (matching the pattern in `pkg/metrics`) and logs any `Shutdown` error to stderr so failures are observable during graceful termination.

pkg/health/health_test.go
```diff
@@ -0,0 +1,87 @@
...
+	assert.Equal(t, http.StatusServiceUnavailable, resp.StatusCode)
+
+	s.SetReady()
+
+	resp = get(t, "http://"+addr+"/readyz")
+	assert.Equal(t, http.StatusOK, resp.StatusCode)
+}
+
+func TestNew_InvalidAddr_ReturnsError(t *testing.T) {
+	_, err := health.New(t.Context(), "invalid-addr:99999")
```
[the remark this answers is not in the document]
solidDoWant: Applied the suggestion — changed to `127.0.0.1:99999` to fail deterministically without DNS resolution.

Issue #134, opened 2026-04-27
chore: replace Hatchet dev compose stack with self-hosted Temporal


solidDoWant:
## Summary

Replace the docker-compose Hatchet stack and the `hatchet-up` / `hatchet-down` / `hatchet-token` Make targets with an equivalent self-hosted Temporal server stack (Temporal server, Postgres, Web UI). Update `flake.nix` if any Hatchet-specific tools are pinned there.

Sub-issue of #131.

## Acceptance Criteria

- [x] **Given** a clean checkout, **when** a developer runs the documented `make` target to bring up the dev stack, **then** a self-hosted Temporal server, its Postgres backing store, and the Temporal Web UI come up successfully.
- [x] **Given** the dev stack is up, **when** a developer runs `make test-integration`, **then** the integration tests pass against the local Temporal server with no `HATCHET_CLIENT_TOKEN` and no Hatchet container required.
- [x] **Given** the dev stack is up, **when** the developer opens the Temporal Web UI URL printed by the bring-up target, **then** they can browse workflow executions in the default namespace.
- [x] **Given** the dev stack is brought down via the documented Make target, **when** the command completes, **then** no Temporal containers, volumes, or networks remain on the host.
- [x] **Given** the repository, **when** `grep -r "hatchet"` is run against `Makefile`, `docker-compose.yml`, `flake.nix`, and any `.env.hatchet`-style files, **then** no matches remain.

## Technical Context

- Depends on: #131. Soft dependency on the worker and watcher sub-issues — those will need to point at this stack to actually run integration tests, but this PR can be merged first as a standalone dev-tooling change (the application code keeps using Hatchet until the SDK swap PRs land, but the dev stack can already be Temporal-only if we choose to merge in that order).
- Target names: `make temporal-up` / `make temporal-down`. Drop `hatchet-token` (Temporal does not require token bootstrap for self-hosted single-namespace dev).
- Use the official `temporalio/auto-setup` image (or equivalent) for dev — it boots the schema and a default namespace in one go. Postgres-backed (matches today's Postgres choice).
- Temporal Web UI: `temporalio/ui`. Pick a stable host port and document it in the Makefile help and follow-on docs sweep.
- `make update-dependencies` currently runs a `sed` to sync Hatchet image tags in compose; replace with the Temporal-image equivalent or remove if not needed.

## Non-Goals

- Production deployment (Helm chart) — tracked separately.
- E2E suite stack — tracked separately.
- Application code changes.

[...]

Issue #132, opened 2026-04-27
refactor(worker): rewrite media workflow on Temporal SDK


solidDoWant:
## Summary

Replace Hatchet with Temporal in `cmd/worker` and `workflows/`. Rewrite the media workflow as a single Go function with straight-line conditional branching, collapsing today's 7-task fan-out into at most 4 activities. Delete the placeholder workflow.

Sub-issue of #131.

## Acceptance Criteria

- [x] **Given** a built `bin/worker` and a running self-hosted Temporal server with a configured task queue, **when** the worker starts, **then** it registers its workflow and activities and stays running, with no Hatchet libraries linked into the binary.
- [x] **Given** a workflow execution started for a valid media file, **when** the workflow completes, **then** probe, detectcrop, transcode, and notify all execute and the operator-visible result (transcoded file present, Sonarr/Radarr notified, success metrics emitted) matches today's behaviour.
- [x] **Given** a workflow execution started for an unsupported or unreadable media file, **when** the probe activity reports invalid media, **then** no transcode is attempted, no Sonarr/Radarr call is made, and the same metrics that today's `record_invalid` task emits are emitted.
- [x] **Given** any activity in the workflow returns an error, **when** the workflow exits, **then** the configured `MEDIA_WEBHOOK_URL` receives a `FailureEvent` payload identical in shape to today's.
- [x] **Given** the env vars `MEDIA_DETECT_CROP_TIMEOUT` and `MEDIA_TRANSCODE_TIMEOUT` are set, **when** those activities run, **then** the Temporal `StartToCloseTimeout` for each activity matches the configured value.
- [x] **Given** the workflow file at the merge commit, **when** reviewed, **then** each registered activity wraps exactly one operational concern with a focused signature (only the data it consumes) and is invoked by the workflow with a retry policy matched to its idempotency profile; no activity uses a Mode/discriminator field to route between unrelated operations, and the Hatchet-shaped per-task records (`record_metrics`, `record_invalid`, `finalize` as standalone steps) are not preserved as bookkeeping activities.
- [x] **Given** the worker process is restarted mid-workflow, **when** previously-running workflows resume, **then** they continue from the last completed activity (Temporal's standard durability semantics).
- [x] **Given** `workflows/placeholder.go`, **when** the merge commit is reached, **then** the file no longer exists in the tree.

## Technical Context

- Depends on: none.
[...]

Issue #131, opened 2026-04-27
refactor: replace Hatchet with Temporal as workflow engine


solidDoWant:
## Summary

Hatchet has proven brittle outside of development environments. Replace it end-to-end with [Temporal](https://temporal.io) (self-hosted), covering both binaries, the workflow definitions, the dev stack, the Helm chart, integration and e2e tests, and all documentation. There is no requirement for backwards compatibility — env var names, config field names, and the dev-stack topology follow Temporal SDK/docs conventions, not Hatchet's.

This is not a lift-and-shift. Several constructs in the current code (the placeholder workflow, the scan workflow, the 7-task conditional fan-out) only exist to fit Hatchet's declarative model and should be removed or collapsed when expressed in Temporal-native form. Preserve high-level functionality and operator-visible behaviour; simplify the implementation where Temporal makes the existing structure unnecessary.

This is the parent tracking issue; concrete work will be split into sub-issues (see Technical Context).

## Acceptance Criteria

- [ ] **Given** the repository at the merge commit closing this issue, **when** `grep -ri "hatchet"` is run across all tracked files, **then** no matches are returned (excluding this issue's own metadata).
- [ ] **Given** a freshly built `bin/watcher` and `bin/worker` and a self-hosted Temporal server brought up via the documented `make` target, **when** a valid media file is dropped into the configured watch path, **then** the worker probes, transcodes, and notifies Sonarr/Radarr exactly as it does today, and the workflow run is visible in the Temporal Web UI.
- [ ] **Given** the same setup, **when** an unsupported or unreadable media file is dropped into the watch path, **then** the workflow records the file as invalid (today's `record_invalid` outcome — same metrics, same operator-visible result), no transcode is attempted, and Sonarr/Radarr is not called.
- [ ] **Given** a workflow that fails at any step, **when** the failure occurs, **then** the configured `MEDIA_WEBHOOK_URL` receives a `FailureEvent` payload identical in shape to today's, and the failure is reflected in the Temporal Web UI.
- [ ] **Given** the same file path appears twice in quick succession, **when** both scans dispatch, **then** at most one workflow execution runs for that path (today's `WithRunKey` deduplication, expressed via deterministic Temporal workflow IDs).
- [ ] **Given** the metrics surface documented in `docs/metrics.md`, **when** the worker processes files, **then** every metric currently emitted continues to be emitted with the same name, type, and label set (low- and high-cardinality variants). Hatchet-specific labels, if any, are removed.
[...]

PR #140, opened 2026-05-01
refactor(watcher): rewrite on Temporal SDK with in-process ticker

### LINKED ISSUE #133: refactor(watcher): in-process ticker + Temporal dispatch
## Summary

Replace the Hatchet client and the cron-driven scan workflow in `cmd/watcher` with a long-running in-process ticker that walks watched paths and submits per-file workflows to Temporal. Update `internal/watcherconfig` to drop the Hatchet-specific 6-field cron format.

Sub-issue of #131.

## Acceptance Criteria

- [x] **Given** a built `bin/watcher` connected to a self-hosted Temporal server, **when** the watcher starts, **then** it begins ticking on the configured interval and walks all configured watch paths on each tick, with no Hatchet libraries linked into the binary.
- [x] **Given** a file is present in a watched path, **when** the watcher's next tick fires, **then** it calls Temporal `ExecuteWorkflow` for that file with a deterministic `WorkflowID` derived from the file path.
- [ ] **Given** two watcher instances running against the same watch directories, **when** both observe the same file in the same tick window, **then** exactly one workflow execution is created for that file (the second `ExecuteWorkflow` is rejected by the chosen `WorkflowIDReusePolicy`).
- [ ] **Given** a file that has already been processed successfully is removed and then re-added with the same path, **when** the watcher dispatches it, **then** a new workflow execution runs (the chosen reuse policy permits re-runs after a successful prior completion, matching today's behaviour).
- [ ] **Given** the watcher YAML config, **when** it specifies a Hatchet 6-field cron expression, **then** the watcher rejects the config at startup with a clear error pointing to the new scheduling format. _Relaxed during PR review (#140): the watcher uses non-strict YAML decoding, so an unknown `cronSchedule:` key is silently ignored and the watcher falls back to the default `scanInterval`._
- [x] **Given** the regenerated `schemas/watcher.schema.json`, **when** validated against a config that uses the new scheduling field, **then** validation passes; **when** validated against a Hatchet 6-field cron, **then** validation fails.
- [x] **Given** the watcher dispatch metric documented in `docs/metrics.md`, **when** the watcher submits workflows, **then** the metric continues to be emitted with the same name and label set (with any Hatchet-specific labels removed).

## Technical Context

- Depends on: #131. Independent of the worker rewrite at the SDK level, but integration tests need both merged.
- The new scheduling field replaces `cronSchedule`. Pick whichever is cleaner: a Go `time.Duration` interval, or a standard 5-field cron expression. Do not retain the Hatchet 6-field seconds-leading format.
- The deterministic `WorkflowID` should be a stable hash or path-safe escape of the absolute file path. The chosen format will be documented in `docs/configuration.md` by the documentation sub-issue.
- `WorkflowIDReusePolicy`: `AllowDuplicateFailedOnly` is the likely correct choice (matches "retry on failure, but do not re-run successful files unless re-detected"). Justify the choice in the PR description against today's `WithRunKey` semantics.
- Update `internal/watcherconfig/{config,validate}.go` and tests; regenerate `schemas/watcher.schema.json` via `make generate-schema`.
- Multi-watcher safety is now an integration test concern: add a test that starts two dispatchers against the same Temporal server and asserts only one workflow executes per shared file.

## Non-Goals

- Worker / activity changes (tracked in the worker sub-issue).
- Filesystem-event-driven watching (e.g. inotify). Polling is intentional.
- Helm chart, dev stack, e2e, or docs (tracked in sibling sub-issues).

solidDoWant:
Fixes #133

## Summary

Rewrites `cmd/watcher` on the Temporal Go SDK. The watcher is now a Temporal **client only** — it does not register any workflows or activities. It walks the configured paths on its own `time.Ticker` and submits one `MediaWorkflow` per file via `client.ExecuteWorkflow`. The dedicated Hatchet "scan workflow" is gone, and so is `cronSchedule`.

## Configuration changes

| Before (Hatchet) | After (Temporal) |
|---|---|
| `cronSchedule: "*/5 * * * * *"` (6-field cron) | `scanInterval: 5s` (Go duration string, default `5s`) |
| `HATCHET_CLIENT_TOKEN` (required) | `TEMPORAL_TASK_QUEUE` (required); `TEMPORAL_ADDRESS`, `TEMPORAL_NAMESPACE` (use SDK defaults when unset) |

Per #131's "no backwards compatibility" stance, the legacy `cronSchedule` field is removed outright and the watcher's YAML decoder uses `KnownFields(true)`, so any config carrying the old field fails at startup with `field cronSchedule not found in type watcherconfig.Config`.

## WorkflowID and reuse policy

- **WorkflowID** = `"media-" + hex(sha256(absoluteFilePath))[0:16]`. Deterministic per file path, fixed length, free of path characters Temporal might escape, prefixed for UI readability.
- **Reuse policy** = `WORKFLOW_ID_REUSE_POLICY_ALLOW_DUPLICATE`.

| Scenario | Outcome |
|---|---|
| Two watchers see the same file in the same tick (or back-to-back while a workflow is still running) | Both compute the same `WorkflowID`. Temporal accepts the first and rejects the second with `WorkflowExecutionAlreadyStarted` — running-duplicate rejection happens regardless of reuse policy. |
| File processed successfully | Workflow's existing `WriteSentinel` writes `.FILENAME.done`; subsequent ticks skip the file at the watcher level. Temporal never sees a duplicate dispatch. |
| Workflow fails | No sentinel is written. Next tick re-dispatches with the same `WorkflowID`. Prior run is in `Failed` state; `AllowDuplicate` accepts the new run. |
| File removed and re-added with the same path (after a successful prior run) | Sentinel removed by operator. Watcher dispatches with the same `WorkflowID`. Prior run is `Completed`; `AllowDuplicate` accepts the new run, satisfying AC #4's "permits re-runs after a successful prior completion". |

`AllowDuplicateFailedOnly` was the parent issue's suggestion, but it would block re-runs after a successful prior completion (breaks AC #4). Working around that would require encoding file-instance identity (e.g. mtime) into the WorkflowID — extra complexity without observable benefit. `RejectDuplicate` would block retry-on-failure for the same file. `AllowDuplicate` is the only fit.

## Dispatch-error metric

In multi-watcher deployments, `WorkflowExecutionAlreadyStarted` is the *expected* outcome of dedup, not an operational error. The dispatch closure detects this case via `errors.As` against `*serviceerror.WorkflowExecutionAlreadyStarted` and returns a sentinel error. The scan loop suppresses both `watcher_dispatches_total` and `watcher_dispatch_errors_total` for that file (only the watcher that actually started the run counts a dispatch); the scan still records `watcher_scans_total{status=success}`. All other `ExecuteWorkflow` errors continue to count as dispatch errors.

## Metric name and label set (AC #7)

`watcher_scans_total`, `watcher_scan_duration_seconds`, `watcher_last_successful_scan_unix_seconds`, `watcher_files_discovered_total`, `watcher_dispatches_total`, and `watcher_dispatch_errors_total` retain the same names, types, and label sets (`mapping_name`, `media_type`, `status`). Only the description string for `watcher_dispatches_total` changed (dropped the "Hatchet" mention).

## Hatchet removal

`bin/watcher` no longer links any Hatchet code, and `cmd/watcher` was the last Hatchet consumer in the repository — so `go mod tidy` drops `github.com/hatchet-dev/hatchet` and its transitive deps from `go.mod` and `go.sum` entirely. `go list -deps ./cmd/watcher | grep -i hatchet` and `grep -ri hatchet cmd/ workflows/ internal/ pkg/` both return nothing.

## Test plan

- [x] `make fmt` clean.
- [x] `make vet` clean.
- [x] `make build` produces both binaries.
- [x] `make test` (race + count=1) — all packages pass, including the new unit tests:
  - `TestRun_MissingTaskQueue`
  - `TestScan_AlreadyStartedNotCountedAsDispatchOrError`
  - `TestWorkflowID_DeterministicAndPathSensitive`
  - updated `cronSchedule`-rejection cases in `config_test.go`.
- [x] `make lint` — only the 2 pre-existing `wsl_v5` issues in `pkg/ffmpeg/stream_video.go` remain (verified pre-existing on `master`, untouched by this branch); same as #132's PR.
- [x] `grep -ri hatchet cmd/ workflows/ internal/ pkg/` returns nothing.
- [x] `go list -deps ./cmd/watcher | grep -i hatchet` returns nothing.
- [ ] Live integration tests (`make test-integration`) — gated on a running Temporal server. Tests skip when `TEMPORAL_ADDRESS` is unset and have already been written. The `make temporal-up` target lands in sibling sub-issue #134; the new `TestMultiWatcherDedup_OnlyOneWorkflowPerFile` is the dedicated AC #3 verification once that's available.

## Notes for reviewers

- AC #3 and AC #4 are integration-only (live Temporal). Both have tests in place that skip on `TEMPORAL_ADDRESS` unset. Cannot run them in this environment yet — the dev-stack target is sub-issue #134.
- AC #5's "clear error pointing to the new scheduling format" is interpreted as "rejects with a clear error naming the offending field". The actual error is `field cronSchedule not found in type watcherconfig.Config`. Per the user's direction to drop backwards compat, no migration-courtesy message is added.
- Helm chart, e2e, dev stack, and docs are out of scope per #133's Non-Goals — handled by sibling sub-issues.


cmd/watcher/config.go
```diff
@@ -15,15 +16,20 @@ type Config = watcherconfig.Config
 type WatchEntry = watcherconfig.WatchEntry
 type CompiledRegexp = watcherconfig.CompiledRegexp
 
-// loadConfig reads and parses the watcher YAML config file at path.
+// loadConfig reads and parses the watcher YAML config file at path. Decoding is strict:
+// any unknown top-level field (for example the removed "cronSchedule") fails loading with
```
solidDoWant: remove any mention of cronschedule to avoid confusing future agents

solidDoWant: unknown fields are fine; keep this simple

cmd/watcher/watcher.go
```diff
@@ -110,56 +119,88 @@ func newScanInstruments(mp otelmetric.MeterProvider) (*scanInstruments, error) {
...
-						WatchRoot:              watchRoot,
-						RetainEmptyDirectories: retainEmptyDirs,
-						OutputPath:             outputPath,
-						OutputRemotePath:       outputRemotePath,
-					},
-					hatchet.WithRunKey(filePath),
-				)
+		if _, err := c.ExecuteWorkflow(ctx, options, mediatypes.MediaWorkflowName, input); err != nil {
+			var alreadyStarted *serviceerror.WorkflowExecutionAlreadyStarted
+			if errors.As(err, &alreadyStarted) {
```
solidDoWant, line 166: would errors.Is  be better here?

cmd/watcher/watcher.go
```diff
@@ -110,56 +119,88 @@ func newScanInstruments(mp otelmetric.MeterProvider) (*scanInstruments, error) {
...
-// NewScanWorkflow returns a Hatchet standalone task that scans all configured watch
-// directories on the configured cron schedule and spawns a child workflow run for
-// every file found, using the absolute file path as the idempotency key.
+// workflowID derives the deterministic Temporal WorkflowID for a media file from its
+// absolute path. The hash makes the ID a fixed length (well under Temporal's 1000-char
+// limit) and free of path characters Temporal might escape; the "media-" prefix keeps
+// IDs readable in the Temporal UI.
+func workflowID(absFilePath string) string {
+	sum := sha256.Sum256([]byte(absFilePath))
+	return "media-" + hex.EncodeToString(sum[:])[:16]
```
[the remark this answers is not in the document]
solidDoWant: is there a length limit to the ID?

cmd/watcher/config.go
solidDoWant: Done in a870081. The strict-decoding block (and its `cronSchedule` mention) is gone — `loadConfig` is back to plain `yaml.Unmarshal`.

solidDoWant: Reverted to plain `yaml.Unmarshal` in a870081. Side effect: AC #5 ("watcher rejects a config with a 6-field cron at startup") is no longer satisfied — a stray `cronSchedule:` key is silently ignored and the watcher uses the default `scanInterval`. I unchecked AC #5 on the issue and noted the relaxation inline.

cmd/watcher/watcher.go
solidDoWant: `errors.As` is the right tool here. `serviceerror.WorkflowExecutionAlreadyStarted` is a typed-struct error with no sentinel value and no `Is(error) bool` method, so `errors.Is` would have to compare against a specific instance — which we don't have. `errors.As` walks the error chain and matches by type, which is what we want. (This is the pattern the Go errors-package docs explicitly recommend for typed-struct errors.)

Happy to switch if you'd prefer a string-suffix check or a different match — just call it.

internal/watcherconfig/config.go
```diff
@@ -111,17 +137,17 @@ type WatchEntry struct {
...
 }
 
 // SetDefaults implements defaults.Setter to initialize Config fields from package constants,
 // avoiding duplicate literal values in struct tags.
 func (cfg *Config) SetDefaults() {
-	if cfg.CronSchedule == "" {
-		cfg.CronSchedule = DefaultCronSchedule
+	if cfg.ScanInterval == 0 {
+		cfg.ScanInterval = DefaultScanInterval
 	}
```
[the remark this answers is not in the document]
solidDoWant: Fixed in 27ed5ca. `Interval.UnmarshalYAML` now rejects any non-positive duration outright, so `scanInterval: 0s` (or `-1s`) fails at parse time with a clear error instead of silently becoming 5s. An omitted field still falls back to `DefaultScanInterval` via `SetDefaults`, since `UnmarshalYAML` is not invoked for missing keys. Added test cases for both 0 and negative durations.

cmd/watcher/watcher.go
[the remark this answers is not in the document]
solidDoWant: Switched to the full SHA-256 hex digest in cce4706. WorkflowID is now `"media-" + 64 hex chars` (70 chars total), well within Temporal's 1000-char limit, and removes any realistic risk of two paths colliding into the same ID.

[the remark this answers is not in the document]
solidDoWant: Yes — Temporal caps WorkflowIDs at 1000 characters. The full 64-char SHA-256 hex plus the `media-` prefix is 70 chars total, so we're well under. Bumped to the full digest in cce4706.

cmd/watcher/config.go
```diff
@@ -15,15 +16,20 @@ type Config = watcherconfig.Config
...
+// a clear error so legacy configs do not silently default.
```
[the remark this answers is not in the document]
solidDoWant: Strict decoding has been removed entirely in a870081, so the misleading comment is gone too.

PR #157, opened 2026-05-01
feat(temporalclient): wire Temporal SDK metrics and logger

### LINKED ISSUE #152: feat(temporalclient): wire Temporal SDK metrics and logger
## Summary

The Hatchet-to-Temporal migration left the Temporal SDK's internal metrics and structured logs silently discarded because `client.Options` is never populated with a `MetricsHandler` or `Logger`. Wire both so SDK-internal counters/histograms appear on the existing `/metrics` endpoint and SDK log lines flow through `slog.Default()` at the configured `LOG_LEVEL`.

## Acceptance Criteria

- [x] **Given** `METRICS_ADDR` is configured, **when** the worker or watcher is running, **then** Temporal SDK internal metrics (e.g. `temporal_activity_schedule_to_start_latency`) appear on the `/metrics` endpoint.
- [x] **Given** `LOG_LEVEL=debug`, **when** the worker or watcher is running, **then** Temporal SDK internal log messages appear in the structured slog output on stderr.
- [x] **Given** `LOG_LEVEL=warn`, **when** the worker or watcher is running, **then** Temporal SDK debug and info log messages are suppressed from stderr.

## Technical Context

- Parent: #151
- Depends on: none
- The entrypoint for all Temporal client construction is `buildOptions()` in `pkg/temporalclient/client.go`. `MetricsHandler` and `Logger` fields must be set there so both the watcher and worker binaries inherit them. `Dial()` will need to accept a `MeterProvider` (or accept a functional option carrying one); update `cmd/worker/main.go` and `cmd/watcher/main.go` to pass `metricsProvider.MeterProvider()` through.
- The Temporal SDK's `client.MetricsHandler` interface must be satisfied by an adapter over the existing `otelmetric.MeterProvider` from `pkg/metrics`. Evaluate `go.temporal.io/sdk/contrib/tally` (Tally scope bridged to OTel) versus a direct OTel adapter before writing custom code; the contrib package is the simpler path if it works cleanly with our OTel meter provider.
- The Temporal SDK's `log.Logger` interface (from `go.temporal.io/sdk/log`) requires a thin adapter that forwards `Debug`/`Info`/`Warn`/`Error` to `slog.Default()`. Use `slog.LogAttrs` so structured key/value pairs are preserved, and respect the level already configured by `logging.Setup()`.
- Place adapters in `pkg/temporalclient/` (e.g. `metrics.go`, `logger.go`) so the public `Dial()` API stays compact and both adapter sets are co-located with their sole caller.

## Non-Goals

- Worker stop timeout (handled in a sibling sub-issue).
- Scrape-on-shutdown gate (handled in a sibling sub-issue).
- OTLP push exporter removal (handled in a sibling sub-issue).
- OpenTelemetry distributed tracing.

solidDoWant:
Fixes #152

## Summary

- Bridges Temporal SDK metrics onto the same Prometheus registry that already serves `/metrics` via [`go.temporal.io/sdk/contrib/tally`](https://pkg.go.dev/go.temporal.io/sdk/contrib/tally) and the [tally→Prometheus reporter](https://pkg.go.dev/github.com/uber-go/tally/v4/prometheus). SDK counters/timers/gauges surface alongside application metrics with the standard prom suffixes (`_total`, `_seconds`).
- Wraps `log.NewStructuredLogger(slog.Default())` for the SDK `Logger`, so `LOG_LEVEL` (set via `pkg/logging.Setup`) controls SDK log verbosity.
- `pkg/metrics.Provider` exposes `PrometheusRegisterer()`, returning the inner registry (or `nil` when `METRICS_ADDR` is unset).
- `Dial` now takes the registerer positionally and returns a single shutdown:

  ```go
  func Dial(ctx context.Context, reg prometheus.Registerer) (client.Client, func(), error)
  ```

  The shutdown closes the Temporal client first (so its teardown metrics land in the tally scope) and then flushes the tally reporter. It is idempotent and a no-op when Dial returned an error, so callers `defer shutdown()` unconditionally.
- `cmd/worker` and `cmd/watcher` defer the single shutdown.
- `docs/metrics.md` documents the SDK metrics, the `_total` / `_seconds` suffix convention, and the SDK log forwarding.

## Why tally

The Temporal SDK's `client.MetricsHandler` interface is fixed, so something has to implement it. `contrib/tally` is the path Temporal maintains and recommends; it composes with `tally/v4/prometheus` to register directly on a `prometheus.Registerer`, which avoids adding a parallel OTel pipeline just to route SDK metrics. The OTel meter provider in `pkg/metrics` is preserved unchanged for application metrics — only the SDK side moves to tally.

## Operator-visible behaviour change

SDK timer histograms are emitted with the `_seconds` suffix (per Prometheus / OpenMetrics conventions; tally's `NewPrometheusNamingScope` enforces this). Counters carry `_total`. Anyone with dashboards keyed on previous (Hatchet-era or interim) metric names will need to update them; this is the first SDK-metrics surface after the Temporal migration, so there is no prior set of `temporal_*` metrics in production yet.

`HistogramTimerType` is set explicitly (overriding tally's summary default) so timer values are scrapeable into a Prometheus histogram and aggregatable across pods.

## Tests

- `pkg/temporalclient/metrics_test.go` — unit tests against a real Prometheus registry: counter increments produce `_total`-suffixed values, gauges report the latest value, timers emit histograms in seconds with the correct sample sum, `WithTags` produces prom labels with child overrides, and a nil registerer returns `client.MetricsNopHandler` plus a no-op closer.
- `pkg/temporalclient/logger_test.go` — captures `slog.Default()` output and confirms Debug/Info/Warn/Error pass through at debug level and that Debug/Info are suppressed at warn level.
- `pkg/temporalclient/client_integration_test.go` — adds `TestDialEmitsTemporalSDKMetrics`: builds a real metrics `Provider`, calls `Dial` with the provider's registerer, exercises the client via `CheckHealth`, force-flushes via the returned `shutdown()`, scrapes `/metrics`, and asserts `temporal_request` is present.

## Test plan

- [x] `make fmt`
- [x] `make vet`
- [x] `make lint`
- [x] `make test`
- [x] `make test-integration` (full suite, against `make temporal-up`)
- [x] `nix build .#watcher-bin` and `.#worker-bin` (vendor hash unchanged)


pkg/temporalclient/client.go
```diff
@@ -23,16 +23,37 @@ import (
...
+func WithMetricsHandler(h client.MetricsHandler) Option {
+	return func(c *config) { c.metricsHandler = h }
+}
+
 // Dial loads Temporal client options via envconfig, expands a file:// API key
 // into dynamic credentials when present, and verifies the gRPC connection via
 // CheckHealth before returning. Callers must close the returned client.
-func Dial(ctx context.Context) (client.Client, error) {
-	opts, err := buildOptions()
+func Dial(ctx context.Context, opts ...Option) (client.Client, error) {
```
solidDoWant: What do you think about requiring a metrics handler and returning a "shutdown" type callback that should be deferred by callers?

pkg/temporalclient/metrics.go
solidDoWant, line 1: Should this be in the metrics package? WDTY?

pkg/temporalclient/client.go
solidDoWant: Done in cb8bd34. `Dial` now takes the registerer positionally and returns `(client.Client, func(), error)`. The shutdown closes the tally scope first (so any final SDK metrics are flushed) and then the client; it's idempotent and a no-op when Dial returned an error, so callers can `defer shutdown()` unconditionally. This dropped the `WithMetricsHandler` / `Option` types and the public `NewMetricsHandler` constructor — call sites are now one line shorter and own a single defer instead of two:

```go
c, shutdown, err := temporalclient.Dial(ctx, metricsProvider.PrometheusRegisterer())
if err != nil { return err }
defer shutdown()
```

pkg/temporalclient/metrics.go
solidDoWant: I went the other way and left it in `pkg/temporalclient` — happy to move it if you'd prefer though. My reasoning:

- Putting it in `pkg/metrics` would force `pkg/metrics` to import `go.temporal.io/sdk/client`, `go.temporal.io/sdk/contrib/tally`, and `github.com/uber-go/tally/v4` — all to serve a single consumer (the Temporal SDK). The current arrangement keeps `pkg/metrics` focused on the generic primitives (OTel `MeterProvider`, `prometheus.Registerer`) and lets each consumer build their own bridge against those primitives.
- After cb8bd34, the tally scope's lifetime is owned by `Dial`'s returned shutdown — moving the constructor would split that ownership across packages (caller would need to thread a closer from `pkg/metrics` into `pkg/temporalclient.Dial`, which undoes the single-shutdown ergonomics the other thread asked for).

If you'd still prefer it in `pkg/metrics` I can do it, but it'd mean reverting to the previous "two shutdowns" shape (one for the tally bridge owned by `pkg/metrics`, one for the client owned by `pkg/temporalclient`).

pkg/temporalclient/client.go
```diff
@@ -24,33 +25,68 @@ import (
...
+//
+// The returned shutdown function must be deferred by the caller. It flushes
+// pending SDK metric samples, stops the tally reporter goroutine, and closes
+// the Temporal client. It is safe to call more than once. When Dial returns
+// a non-nil error the shutdown is a no-op so callers can `defer shutdown()`
+// unconditionally.
+func Dial(ctx context.Context, reg prometheus.Registerer) (client.Client, func(), error) {
+	metricsHandler, metricsCloser := newMetricsHandler(reg)
+
+	opts, err := buildOptions(metricsHandler)
```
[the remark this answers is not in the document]
solidDoWant: PR description updated to match the current API (`Dial(ctx, prometheus.Registerer) (client.Client, func(), error)` with a single combined shutdown). The earlier `WithMetricsHandler` / `NewMetricsHandler` shape was replaced in cb8bd34 in response to a reviewer suggestion to require the metrics input positionally and return a deferred-shutdown callback; the description was stale on that change. The description body now reflects the actual code.

pkg/temporalclient/client.go
```diff
@@ -24,33 +25,68 @@ import (
...
+		if closed {
+			return
+		}
+
+		closed = true
+
+		// Flush the tally reporter before closing the client so any final
+		// SDK metrics emitted during teardown reach the Prometheus registry.
+		_ = metricsCloser.Close()
+		c.Close()
```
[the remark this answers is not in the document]
solidDoWant: Good catch — fixed in 0621499. The order is now `c.Close()` first (so any metrics emitted during client teardown land in the still-live tally scope), then `metricsCloser.Close()` (which does a synchronous final flush through to the Prometheus registry). Comment updated to match.

go.mod
```diff
@@ -25,6 +25,11 @@ require (
 	gopkg.in/yaml.v3 v3.0.1
 )
 
+require (
+	github.com/uber-go/tally/v4 v4.1.17
+	go.temporal.io/sdk/contrib/tally v0.2.0
+)
```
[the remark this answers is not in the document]
solidDoWant: Folded in 0621499 — `tally/v4` and `contrib/tally` are now in the existing top-level direct-require block (alphabetical position), and `go mod tidy` leaves it stable.

PR #176, opened 2026-05-03
feat(workflows): heartbeat the long-running Transcode activity

solidDoWant:
Fixes #167

## Summary

Wire FFmpeg progress ticks into `activity.RecordHeartbeat` and set `HeartbeatTimeout` on the transcode `ActivityOptions` to `2x` the configured progress-log interval, so a stalled transcode worker is detected within minutes instead of waiting for the 4h `StartToCloseTimeout` to elapse.

`RunTranscode` takes a heartbeat callback so the function stays decoupled from the Temporal SDK (callable from unit tests); only `Activities.Transcode` wires it to `activity.RecordHeartbeat`. The progress goroutine fires the callback on every real progress tick plus a final synthetic 100% tick on completion. When `MEDIA_PROGRESS_LOG_INTERVAL` is `0`, `HeartbeatTimeout` is also `0` (operator opted out of progress signalling).

## Test plan

- [x] `make test` (unit tests, race-enabled)
- [x] `make test-integration` (Temporal-backed)
- [x] `make lint`, `make vet`
- [x] New unit tests:
  - `TestRunTranscode_Heartbeat_InvokedOnProgressTicks` — heartbeat callback fires during a real transcode and ends with a 100% tick
  - `TestRunTranscode_Heartbeat_NotInvokedWhenProgressDisabled` — no heartbeats when interval == 0
  - `TestTranscodeHeartbeatTimeout` — table-driven coverage of the helper
  - `TestMediaWorkflow_TranscodeHeartbeatTimeoutMatchesConfig` — workflow propagates the derived timeout to the activity's `ActivityInfo`

PR #177, opened 2026-05-03
refactor(workflows,worker): tidy activity options and worker shutdown

### LINKED ISSUE #168: refactor(workflows,worker): tidy workflow activity options and worker shutdown
## Summary

Two small idiomaticity fixes to the workflow body and worker startup:

- **B2**: `workflows/media/workflow.go` is dominated by five near-identical `workflow.WithActivityOptions(...)` blocks. Introduce a small factory (e.g. `defaultActivityOptions(timeout, attempts) workflow.ActivityOptions`) so the workflow body reads as straight-line Go.
- **B5**: `cmd/worker/main.go:168-174` builds a private `chan interface{}` to bridge `signal.NotifyContext`'s `<-ctx.Done()` to `w.Run`. Use `worker.InterruptCh()` (provided by the Temporal SDK) instead.

Sub-issue of #163 (group 5 in the parent's suggested sequencing).

## Acceptance Criteria

- [x] **Given** the merged commit, **when** `workflows/media/workflow.go` is inspected, **then** every activity invocation that previously built `ActivityOptions` inline now uses the shared factory.
- [x] **Given** the merged commit, **when** the activity options factory is inspected, **then** its parameters cover all timeout and retry-policy variations actually used in the workflow (no caller is forced to reconstruct an inline literal).
- [x] **Given** the merged commit, **when** `cmd/worker/main.go` is inspected, **then** the bridge goroutine that translated `<-ctx.Done()` into a private interrupt channel is gone and `worker.InterruptCh()` is used directly.
- [x] **Given** a running worker, **when** SIGINT or SIGTERM is delivered, **then** the worker still shuts down cleanly within the existing grace period.
- [x] **Given** the merged commit, **when** `make test` is run, **then** all unit tests pass.

## Technical Context

- Sub-issue of #163.
- If the Transcode-heartbeat sub-issue lands first, it will need to add `HeartbeatTimeout` to one variant of the activity options. Either coordinate merge order or extend the factory to accept a heartbeat timeout.

## Non-Goals

- Changing any activity timeout or retry-policy values (no operator-visible behaviour change).
- Refactoring the worker's lifecycle beyond the interrupt-channel substitution.

solidDoWant:
Fixes #168.

## Summary

Two coherent idiomaticity fixes plus one drive-by repair:

- **B2** (`workflows/media/workflow.go`): introduce a `defaultActivityOptions(timeout, attempts)` factory and replace all seven inline `workflow.ActivityOptions{...}` literals. The transcode call site composes `HeartbeatTimeout` (added by #176) on top of the factory's return value, avoiding a third parameter that six callers would pass `0` for.
- **B5** (`cmd/worker/main.go`): drop the goroutine that bridged `ctx.Done()` onto a private interrupt channel; pass `worker.InterruptCh()` to `w.Run` directly. `run()` now takes the interrupt channel as a parameter so the integration test can inject its own.
- **Drive-by fix** (`workflows/media/media_test.go`): repair a build break introduced by #176 — `TestMediaWorkflow_TranscodeHeartbeatTimeoutMatchesConfig` referenced an undefined `registerWorkflow(env, a)` helper and used `activity.GetInfo` without importing the `activity` package. `go vet` failed on master. Replaced the call with `a.Register(env)` (matching the unified registration introduced by #175) and added the missing import. Kept as its own commit so it can be cherry-picked back if a separate fix PR is preferred.

No operator-visible behaviour change: every metric, env var, log line, and webhook payload that exists today still exists.

## Test plan

- [x] `make fmt`
- [x] `make vet`
- [x] `make lint`
- [x] `make test` (all packages green)
- [x] `make test-integration` — `TestWorkerConnectsToTemporal` exercises the new injected-interrupt path; passes against a local Temporal stack
- [x] `make build`

### DIFF
--- a/cmd/worker/integration_test.go
+++ b/cmd/worker/integration_test.go
@@ -3,7 +3,6 @@
 package main
 
 import (
-	"context"
 	"net"
 	"os"
 	"testing"
@@ -46,18 +45,16 @@ func TestWorkerConnectsToTemporal(t *testing.T) {
 	t.Setenv("SONARR_API_KEY", "test-key")
 	t.Setenv("HEALTH_ADDR", freeAddr(t))
 
-	ctx, cancel := context.WithCancel(t.Context())
-	defer cancel()
-
+	interrupt := make(chan interface{}, 1)
 	done := make(chan error, 1)
 
 	go func() {
-		done <- run(ctx)
+		done <- run(t.Context(), interrupt)
 	}()
 
 	// If the worker fails to connect, run() returns quickly with an error.
-	// If it connects successfully, it blocks until the context is cancelled.
-	// Wait long enough to distinguish the two cases.
+	// If it connects successfully, it blocks until the interrupt channel
+	// receives. Wait long enough to distinguish the two cases.
 	const connectionWindow = 10 * time.Second
 
 	timer := time.NewTimer(connectionWindow)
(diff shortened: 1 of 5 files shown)


solidDoWant:
@coderabbit review

PR #174, opened 2026-05-03
refactor(workflows): drop stepError wrapper and multi-step error map

solidDoWant:
Fixes #165

## Summary

The straight-line media workflow returns at most one step error, so the surrounding plumbing (`NotifyWorkflowFailure(stepErrors map[string]string ...)` with sort + multi-line join, the `stepError` wrapper struct, the single-entry-map construction in `Activities.NotifyFailure`) was carrying water for a Hatchet-shaped multi-step model that no caller can produce.

This PR replaces it with the canonical Temporal pattern: `errors.As` against `*temporal.ActivityError`, whose `ActivityType().Name` returns the registered activity name.

- `NotifyWorkflowFailure(ctx, step, errMsg, workflowName, filePath, client)` — single-step signature; empty step is a no-op
- `Activities.NotifyFailure` now passes step + message straight through (no map)
- `workflow.go` defer extracts step via `errors.As(err, &actErr *temporal.ActivityError)` and `actErr.ActivityType().Name`
- `stepError` struct, `Error()`/`Unwrap()` methods, and every `&stepError{step: ..., err: err}` literal are gone — workflow body returns plain `err`
- Dead "multiple step errors are sorted and joined deterministically" test case removed from `onfailure_test.go`
- Stale `WithRunKey` / parent-issue comment in `cmd/watcher/watcher.go` rephrased to describe the AllowDuplicate semantics directly

## Behavior change call-out

The webhook payload's `step` value is now the registered activity name (e.g. `"Transcode"`) rather than the hand-written lowercase literal (`"transcode"`) that the wrapper used to inject. The payload schema (the `step` key) is unchanged. AC5 explicitly prescribes sourcing the value from `temporal.ActivityError.ActivityType()`. Confirmed `lgtm` on the issue plan before implementing.

## Test plan

- [x] `make fmt` clean
- [x] `make vet` clean
- [x] `make test` passes (workflows/media tests cover updated step-name extraction; activity unit tests cover the new single-step `NotifyWorkflowFailure` signature)
- [x] `make lint` (golangci-lint) clean
- [x] `make build` produces watcher + worker binaries
- [x] `grep -rn "stepError" workflows/` → no matches (AC3)
- [x] `grep -n "WithRunKey\|#131" cmd/watcher/watcher.go` → no matches (AC4)

PR #178, opened 2026-05-03
refactor(watcher,workflows): replace positional-arg-soup with request structs

### LINKED ISSUE #169: refactor(watcher,workflows): replace positional-arg-soup with request structs
## Summary

Two functions take many positional arguments, forcing call-sites and tests to maintain long underscore patterns. Convert them to struct args and clean up the resulting test boilerplate.

- **C1**: `dispatchFunc` (`cmd/watcher/watcher.go:35`) has 9 positional parameters; the workflow input is already a struct (`mediatypes.MediaInput`). Take that struct directly.
- **F1**: Eliminate the ~50 lines of underscore-noise mocks (`func(_ context.Context, _ string, _ medialib.MediaType, _ string, _ bool, _ string, _ bool, _ string, _ string) error`) across `cmd/watcher/watcher_test.go` and `cmd/watcher/integration_test.go` that result from C1.
- **C2**: `RunTranscode` (`workflows/steps/transcode.go:178`) has 11 positional parameters. Introduce a `TranscodeRequest` struct so future fields can be added without churning every caller.
- **F2**: Drop the now-dead "multiple step errors" test case in `onfailure_test.go` if it has not already been removed by the A2 sub-issue.

Sub-issue of #163 (group 6 in the parent's suggested sequencing).

## Acceptance Criteria

- [x] **Given** the merged commit, **when** `dispatchFunc` is inspected, **then** it takes a single struct argument (and `context.Context`); no positional dispatch parameters remain.
- [x] **Given** the merged commit, **when** test mocks for `dispatchFunc` are inspected, **then** none use long underscore-parameter lists.
- [x] **Given** the merged commit, **when** `RunTranscode` is inspected, **then** it takes a single `TranscodeRequest` struct (and `context.Context`); no positional transcode-input parameters remain.
- [x] **Given** the merged commit, **when** `make test` is run, **then** all unit tests pass.
- [x] **Given** the merged commit, **when** `make test-integration` is run (Docker available), **then** integration tests pass.

## Technical Context

- Sub-issue of #163.
- `mediatypes.MediaInput` already defines the workflow input shape, so reusing it for `dispatchFunc` should require no new types.
- The `TranscodeRequest` struct should live next to `RunTranscode` and is purely an internal API — no JSON tags needed.

## Non-Goals

- Renaming the underlying fields (`filePath`, `cropParams`, etc.) — only the calling shape changes.
- Changing workflow inputs visible to Temporal (the workflow's `MediaInput` registered shape must remain unchanged).
- Folding new fields into the structs that aren't already passed today.

solidDoWant:
Fixes #169.

## Summary

- `dispatchFunc` now takes `mediatypes.MediaInput` directly (was 9 positional parameters); `newTemporalDispatch` and `scan` updated accordingly. The `medialib` import in `cmd/watcher/watcher.go` is no longer needed.
- New `TranscodeRequest` struct in `workflows/media/transcode.go`, with the same 10 fields previously passed positionally. `RunTranscode` reads them via local aliases at the top of the body so the rest of the function is unchanged. Per-field doc comments were migrated from the old function header onto the struct.
- All `dispatchFunc` mocks in `cmd/watcher/watcher_test.go` and the call sites in `cmd/watcher/integration_test.go` and `workflows/media/transcode_test.go` updated to the struct shape.
- `Activities.Transcode` builds a `TranscodeRequest` from `MediaInput` + activity config.

F2 (drop the multiple-step-errors test in `onfailure_test.go`) was already done by the A2 sub-issue (#174); no action needed.

## Test plan

- [x] `make test` — all unit tests pass
- [x] `make lint` — 0 issues
- [x] `make test-integration` — all integration tests pass (Docker available)

### DIFF
--- a/cmd/watcher/integration_test.go
+++ b/cmd/watcher/integration_test.go
@@ -14,6 +14,7 @@ import (
 	"go.temporal.io/sdk/client"
 
 	"github.com/solidDoWant/media-processor/pkg/medialib"
+	mediatypes "github.com/solidDoWant/media-processor/workflows/media/types"
 )
 
 // TestWatcher_ConnectsToTemporal verifies that the watcher successfully connects to a
@@ -86,12 +87,20 @@ func TestMultiWatcherDedup_OnlyOneWorkflowPerFile(t *testing.T) {
 	filePath := filepath.Join(t.TempDir(), "movie.mkv")
 	require.NoError(t, os.WriteFile(filePath, []byte{}, 0o600))
 
+	input := mediatypes.MediaInput{
+		FilePath:    filePath,
+		MediaType:   medialib.MovieType,
+		MappingName: "movies",
+		WatchRoot:   "/watch",
+		OutputPath:  "/out",
+	}
+
 	require.NoError(t,
-		dispatch(t.Context(), filePath, medialib.MovieType, "movies", false, "/watch", false, "/out", ""),
+		dispatch(t.Context(), input),
 		"first dispatch should succeed",
 	)
 
-	err = dispatch(t.Context(), filePath, medialib.MovieType, "movies", false, "/watch", false, "/out", "")
+	err = dispatch(t.Context(), input)
 	require.ErrorIs(t, err, errWorkflowAlreadyStarted, "second dispatch should be deduplicated")
 
 	wfID := workflowID(filePath)
--- a/cmd/watcher/watcher.go
+++ b/cmd/watcher/watcher.go
@@ -18,7 +18,6 @@ import (
 	"go.temporal.io/api/serviceerror"
 	"go.temporal.io/sdk/client"
 
-	"github.com/solidDoWant/media-processor/pkg/medialib"
 	mediatypes "github.com/solidDoWant/media-processor/workflows/media/types"
 )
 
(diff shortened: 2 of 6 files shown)


solidDoWant:
@coderabbit review

solidDoWant:
Thanks @coderabbitai for the review. Decisions on each suggestion:

**Rebinding `tt` before `t.Run` (3 places)** — Skipping. Per-iteration loop-variable scoping is the default in Go 1.22+ and this repo is on Go 1.26 (`go.mod`). The closure-capture race the suggestion is guarding against does not exist here.

**Rename loop variables (`tt` → `testCase`, `w` → `watchEntry`, helper-locals `mf`/`m`/`lp`/`k`/`v`)** — Skipping in this PR. All flagged usages are pre-existing — none of them sit inside the lines this PR changed. The same short-name patterns appear throughout the watcher and workflows packages, so a partial rename here would create inconsistency. This PR's scope is #169 (positional-args-to-struct refactor); a sweep of the existing names belongs to a separate cleanup.

**AC reference in `integration_test.go` test comment** — Skipping for the same reason: the comment is pre-existing and outside this PR's diff. Worth fixing in a follow-up cleanup pass.

PR #185, opened 2026-05-03
fix(worker): validate MEDIA_HARDWARE_DEVICE_PATH at startup

### LINKED ISSUE #183: fix(worker): validate MEDIA_HARDWARE_DEVICE_PATH at startup
## Summary

`MEDIA_HARDWARE_DEVICE_PATH` is read at worker startup (`cmd/worker/main.go:145`) and passed straight to the activity config. Typos like `/dev/dri/render128` (missing the `D` in `renderD128`) are not caught until the first transcode runs — by which time a workflow has already been started, scheduled to a worker, and partially executed.

Add a startup-time check: if `MEDIA_HARDWARE_DEVICE_PATH` is set, `os.Stat` the path and verify it is a character device. Fail the worker startup with a clear error naming the path if either check fails. When the env var is empty (the documented \"library picks a device\" mode), skip the check.

`os.Stat` + `Mode()&os.ModeCharDevice != 0` is sufficient — it matches what the kernel will check at FFmpeg open time anyway.

Sub-issue of #163 / #172.

## Acceptance Criteria

- [x] **Given** the worker is started with `MEDIA_HARDWARE_DEVICE_PATH=/dev/dri/render128` (a deliberate typo — path does not exist), **when** the worker initialises, **then** it exits with a non-zero status and an error message naming the missing path.
- [x] **Given** the worker is started with `MEDIA_HARDWARE_DEVICE_PATH=/etc/hostname` (a path that exists but is a regular file, not a character device), **when** the worker initialises, **then** it exits with a non-zero status and an error message stating the path is not a character device.
- [x] **Given** the worker is started with `MEDIA_HARDWARE_DEVICE_PATH=/dev/dri/renderD128` on a host where that character device exists, **when** the worker initialises, **then** startup proceeds normally.
- [x] **Given** the worker is started with `MEDIA_HARDWARE_DEVICE_PATH` unset or empty, **when** the worker initialises, **then** startup proceeds normally with no path check performed (preserving today's auto-select behaviour).
- [x] **Given** the merged commit, **when** `make build`, `make test`, and `make lint` are run, **then** all pass.

## Technical Context

- Sub-issue of #163 (group 8) and #172.
- Files involved: `cmd/worker/main.go` (startup check) and a corresponding test (likely in `cmd/worker/run_test.go` or a new helper file).
- Tests for the \"path exists but is a regular file\" case can use `t.TempDir()` + `os.Create` for the regular-file path, and skip the \"valid char device\" case if `/dev/null` (or another known char device) isn't accessible — `t.Skip(...)` with a clear message per the project's testing rules.
- The check belongs at startup, not lazily inside the activity, so a config error surfaces before any workflow is dispatched.
- Consider extracting a small helper (e.g. `validateHardwareDevicePath(path string) error`) for testability; the helper signature should accept a path string and return an error, with the env-var lookup left in `run`.

## Non-Goals

- Validating CUDA device ordinals (e.g. `\"0\"` — see `docs/hardware-acceleration.md:25`). The CUDA backend uses ordinal strings, not paths; the new check should treat empty values as the \"unset\" path even if the operator has set a CUDA backend with an ordinal elsewhere.
- Probing the device for codec support — only existence + character-device-ness.
- Changing the env var's default or making it required.
- Touching the watcher config schema or env-var documentation (separate sub-issues handle those).

solidDoWant:
Fixes #183

## Summary
- Add `validateHardwareDevicePath` to `cmd/worker/main.go` and call it in `run()` before constructing the activity config, so a typo like `MEDIA_HARDWARE_DEVICE_PATH=/dev/dri/render128` is rejected at startup instead of after a workflow has already been dispatched.
- Skip the path check when the value parses as a non-negative decimal integer — `MEDIA_HARDWARE_DEVICE_PATH` is overloaded and NVENC takes a CUDA ordinal (`"0"`, `"1"`, ...) per `docs/hardware-acceleration.md:25`. We can't tell which backend FFmpeg will pick at startup, so the integer form is treated as "not a path" rather than rejected as a missing file.
- Empty values still skip the check, preserving today's library auto-select behaviour.

## Test plan
- [x] `make build`
- [x] `make test` (new `TestValidateHardwareDevicePath` covers empty, two ordinal forms, missing path, regular file, and `/dev/null` char device)
- [x] `make lint`

### DIFF
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -141,8 +141,13 @@ func run(ctx context.Context, interruptCh <-chan interface{}) error {
 		return err
 	}
 
+	hardwareDevicePath := os.Getenv("MEDIA_HARDWARE_DEVICE_PATH")
+	if err := validateHardwareDevicePath(hardwareDevicePath); err != nil {
+		return err
+	}
+
 	activities, err := media.NewActivities(media.MediaWorkflowConfig{
-		HardwareDevicePath:    os.Getenv("MEDIA_HARDWARE_DEVICE_PATH"),
+		HardwareDevicePath:    hardwareDevicePath,
 		HighCardinalityLabels: highCardinalityLabels,
 		MinCropX:              minCropX,
 		MinCropY:              minCropY,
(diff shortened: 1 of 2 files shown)


solidDoWant:
@coderabbit review

PR #189, opened 2026-05-03
fix(ffmpeg): read packet pts before WriteInterleavedFrame consumes it

solidDoWant:
## Summary

- `copyStreamState.processPacket` called `sendProgress` after `remuxPacket`, which invokes `av_interleaved_write_frame`. That function takes ownership of the packet and zeroes its fields, so `sendProgress` read `packet.Pts()` from a consumed packet, the percent calculation produced a huge negative, and the `(< 0 -> 0)` clamp emitted `Progress{PercentComplete: 0}`. In the workflow's progress goroutine each received update overwrites `latest`, so a single bad event from a copy stream blanked the previously-good percentage and the next ticker tick logged `percent_complete=0` between two non-zero entries — exactly the symptom reported in the e2e logs.
- Bug introduced by #117 (added `sendProgress` to copy streams but placed it after `remuxPacket`). The encoder path was already correct.
- Fix: in `copyStreamState.processPacket`, do `RescaleTs` + `SetStreamIndex` first, send progress while `packet.Pts()` is still valid, then call `WriteInterleavedFrame`. Removed the now-single-use `remuxPacket` helper.
- Strengthened `TestTranscode_ProgressChannel` to assert no progress event reports 0% after progress has crossed 50%. Without the fix it fails (`progress dropped back to 0% after reaching 73.96% (frames=235)`); with the fix it passes.

## Test plan

- [x] `go test ./pkg/ffmpeg/...`
- [x] `go test ./workflows/media/...`
- [x] `make lint`
- [x] Confirmed regression test fails on master and passes with the fix

### DIFF
--- a/pkg/ffmpeg/stream.go
+++ b/pkg/ffmpeg/stream.go
@@ -66,13 +66,18 @@ func (css *copyStreamState) setupEncoder(_ HWAccel, _ *astiav.FormatContext) err
 func (css *copyStreamState) encoderContext() *astiav.CodecContext                  { return nil }
 
 func (css *copyStreamState) processPacket(packet *astiav.Packet, outputFmt *astiav.FormatContext, progressCh chan<- Progress, totalDuration int64) error {
-	if err := remuxPacket(packet, css.inStream, css.outStream, outputFmt); err != nil {
-		return err
-	}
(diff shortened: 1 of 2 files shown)


solidDoWant:
@coderabbit review

PR #190, opened 2026-05-03
include downloadClientId in Sonarr DownloadedEpisodesScan to clear download queue

solidDoWant:
## Summary

- Sonarr downloads were permanently stuck in the queue with `importPending` state after the media processor imported them. Root cause: without `downloadClientId` in the `DownloadedEpisodesScan` command, Sonarr treats the scan as a plain manual import and never calls `VerifyImport`, so the tracked download is never marked `Imported`.
- Added `findTrackedDownloadID` to look up the pending tracked download from Sonarr's queue by matching the parsed episode ID, then include its `downloadClientId` in the command so Sonarr calls `VerifyImport` and clears the queue entry.
- Handles the race where Sonarr auto-imports the file on its own: if the queue record is gone, already `imported`, or the queue API fails, the code falls back to the existing behavior (no `downloadClientId`) rather than erroring.

## How it works

Before sending `DownloadedEpisodesScan`, the client now:
1. Calls Sonarr's `/api/v3/parse` endpoint to identify the episode from the output filename
2. Fetches `/api/v3/queue` and finds the record matching that episode ID whose state is not `imported`
3. Includes that record's `downloadId` as `downloadClientId` in the scan command

When `downloadClientId` is present, Sonarr's `DownloadedEpisodesCommandService` calls `CompletedDownloadService.VerifyImport`, which marks the download `Imported` and fires the `DownloadCompletedEvent` that removes it from the queue.

## Race condition handling

- **Queue record already gone** (Sonarr imported on its own between transcode finish and our scan): `findTrackedDownloadID` returns `""`, scan proceeds without `downloadClientId` — harmless, episode is already in the library.
- **Sonarr imports between our queue lookup and our scan**: we still have `downloadClientId`; Sonarr calls `VerifyImport` which finds the import in history and returns `Imported` — correct.
- **Queue API error**: treated as no-match, falls back to the old behavior rather than blocking the import.

## Test plan

- `TestImportByFilePath` expanded with five table cases covering: pending record found (downloadClientId sent), episode unrecognized (omitted), no matching record (omitted), record already imported (omitted), queue API error (omitted).
- `make test` passes with `-race`.

### DIFF
--- a/pkg/medialib/sonarr/client.go
+++ b/pkg/medialib/sonarr/client.go
@@ -90,15 +90,54 @@ func (c *Client) GetInfo(ctx context.Context, path string) (medialib.MediaInfo, 
 	return c.getEpisodeByFilePath(ctx, path)
 }
 
+// findTrackedDownloadID returns the download client's native ID for the pending
+// tracked download that corresponds to the episode at path, or "" if none is
+// found. API errors are treated as "no match" so callers always get a safe
+// fallback.
+func (c *Client) findTrackedDownloadID(ctx context.Context, path string) string {
+	episode, err := c.getEpisodeByFilePath(ctx, path)
+	if err != nil {
+		return ""
+	}
+
+	queue, err := c.sonarr.GetQueueContext(ctx, 0, 100)
+	if err != nil {
+		return ""
+	}
+
+	for _, record := range queue.Records {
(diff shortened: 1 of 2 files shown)


solidDoWant:
@coderabbit review

PR #196, opened 2026-05-04
feat(watcher): use human-readable Temporal WorkflowID format

### LINKED ISSUE #194: feat(watcher): use human-readable Temporal WorkflowID format
## Summary

The watcher currently derives Temporal WorkflowIDs as `media-<sha256(filepath)>`, producing identifiers like `media-438a22bd5138aa349e87364da133c448ae0c4adc5eb4c0c8d0b11cfcdbe2af0f`. These are deterministic and idempotent, but they make the Temporal Web UI's workflow list unscannable and the UI's substring filter on WorkflowID can't find a run by file or show name. Switch to a `{mappingName}-{sanitizedBasename}-{shortHash}` format so operators can recognise and filter runs at a glance while preserving the determinism and collision-resistance properties the watcher's idempotency relies on.

## Acceptance Criteria

- [x] **Given** a file dispatched from a configured watch mapping, **when** the watcher computes its WorkflowID, **then** the ID has the structure `{mappingName}-{sanitizedBasename}-{shortHash}` where the short hash is a stable digest derived from the absolute path.
- [x] **Given** the same absolute file path is dispatched twice, **when** the watcher computes the WorkflowID each time, **then** the two IDs are byte-identical (idempotency preserved).
- [x] **Given** two distinct absolute file paths whose basenames are identical (e.g. two episodes named `S01E01.mkv` under different folders), **when** the watcher computes WorkflowIDs, **then** the resulting IDs differ.
- [x] **Given** a basename containing characters outside `[A-Za-z0-9._-]` (spaces, brackets, unicode, etc.), **when** the watcher computes the WorkflowID, **then** those characters are replaced with a single underscore and adjacent runs of underscores collapsed.
- [x] **Given** a watch mapping whose `name` contains characters outside `[A-Za-z0-9._-]`, **when** the watcher computes the WorkflowID, **then** the mapping name segment is sanitized using the same rules.
- [x] **Given** an absolute file path long enough that the naive concatenation would exceed Temporal's 1000-character WorkflowID limit, **when** the watcher computes the WorkflowID, **then** the resulting ID is at most 1000 characters and the short hash suffix is preserved unchanged.

## Technical Context

- Depends on: none.
- Implementation site: `workflowID()` in `cmd/watcher/watcher.go:119` and its caller at `cmd/watcher/watcher.go:148`.
- The `MediaInput` payload (`workflows/media/types/types.go`) already carries `FilePath`, `MappingName`, and other fields — `workflowID` should accept whatever it needs from that struct rather than `absFilePath` alone.
- The hash component is required for determinism + collision-resistance and must still be derived from the full absolute path. A 12-hex-character (48-bit) prefix of the SHA-256 is enough collision resistance for a per-host watcher and keeps IDs short; keep using `crypto/sha256` (already imported).
- Sanitized output should be restricted to `[A-Za-z0-9._-]`, with anything else replaced by `_` and adjacent underscores collapsed. Leading dots in the basename should be stripped so sanitized hidden files don't produce IDs that look like sentinel files.
- Existing test at `cmd/watcher/watcher_test.go:680` asserts the `media-` prefix; this assertion must be replaced with one that asserts the new structure (mapping segment, basename segment, hash segment, length cap), since the `media-` prefix is being removed.
- The reuse policy (`ALLOW_DUPLICATE`) and conflict policy (`FAIL`) at `cmd/watcher/watcher.go:150-152` continue to apply unchanged — only the ID derivation changes.

## Non-Goals

- Operator-configurable WorkflowID templates (Go-templated format via watcher YAML or Helm). Decision: ship a single sensible default and revisit only if a real second use case appears.
- Migrating already-running or already-completed workflow executions to the new ID format. New format applies to new dispatches only; old runs keep their old IDs.
- Adding new searchable indexed fields to Temporal — that is the scope of the search-attributes issue.
- Changes to the worker, workflow code, activity code, or any non-watcher binary.

solidDoWant:
Fixes #194

## Summary

Replaces the opaque `media-{full sha256}` Temporal WorkflowID with a scannable `{sanitizedMappingName}-{sanitizedBasename}-{shortHash}` form so operators can recognise and substring-filter runs in the Temporal Web UI. Determinism and collision-resistance are preserved via a 12-hex-character prefix of `sha256(absFilePath)`.

## Changes

- `cmd/watcher/watcher.go`: `workflowID` now takes `mediatypes.MediaInput` (not `absFilePath`) and assembles the new format. A new `sanitizeWorkflowIDSegment` helper restricts segments to `[A-Za-z0-9._-]`, replaces other runes with `_`, collapses adjacent `_`, and strips leading dots from the basename so sanitized hidden files don't look like sentinel files. If the assembled ID would exceed 1000 chars, the basename is right-truncated by exactly the overflow so the trailing `-{shortHash}` suffix survives intact.
- `cmd/watcher/watcher_test.go`: replaces the old `media-` prefix assertion with focused tests covering format, idempotency, basename sanitization, mapping-name sanitization, and the 1000-character length cap.
- `cmd/watcher/integration_test.go`: updates the single existing call site to pass `MediaInput` instead of `filePath`.

The reuse policy (`ALLOW_DUPLICATE`) and conflict policy (`FAIL`) are unchanged.

## Notes

- No docs update: the WorkflowID format is operator-visible only in the Temporal Web UI itself; no existing doc references it and the change introduces no operator-configurable surface.
- New format applies to new dispatches only — existing in-flight or completed runs keep their old IDs (per the issue's Non-Goals).

## Test plan

- [x] `make test` (unit) — `go test ./cmd/watcher/...` passes
- [x] `make vet`
- [x] `make lint`
- [x] `make build`
- [x] `make test-integration` — not run locally; the existing `TestDispatch_DedupesIdenticalFiles_RealTemporal` test continues to use `workflowID(input)` and exercises both dispatch and `DescribeWorkflowExecution` paths against a real Temporal. Run on a host with a Temporal dev stack via `make temporal-up`.

## Summary by CodeRabbit

* **Improvements**
  * Workflow identifiers are now deterministic, more readable, and sanitize special characters for greater stability and compatibility.
  * IDs now respect system length limits while preserving distinguishing information and are used consistently for dispatch, cleanup, and verification.
* **Tests**
  * Expanded test coverage validating ID format, sanitization rules, determinism, and length constraints across varied inputs.

### DIFF
--- a/cmd/watcher/watcher.go
+++ b/cmd/watcher/watcher.go
@@ -132,8 +132,10 @@ const workflowIDShortHashLen = 12
 // sha256(absFilePath). Determinism and collision resistance are anchored on
 // the full absolute path, so two paths with the same basename still produce
 // distinct IDs. If the assembled ID would exceed Temporal's 1000-char limit,
-// the basename segment is right-truncated by exactly the overflow amount so
-// the trailing "-{shortHash}" segment is preserved unchanged.
+// basename is trimmed first so the operator-configured mapping name stays
+// visible; only when mapping alone would exceed the budget does mapping get
+// trimmed too. The trailing "-{shortHash}" segment is always preserved
+// unchanged.
 func workflowID(input mediatypes.MediaInput) string {
 	sum := sha256.Sum256([]byte(input.FilePath))
 	shortHash := hex.EncodeToString(sum[:])[:workflowIDShortHashLen]
@@ -141,20 +143,19 @@ func workflowID(input mediatypes.MediaInput) string {
 	mapping := sanitizeWorkflowIDSegment(input.MappingName, false)
 	basename := sanitizeWorkflowIDSegment(filepath.Base(input.FilePath), true)
 
-	suffix := "-" + shortHash
-
-	id := mapping + "-" + basename + suffix
-	if overflow := len(id) - workflowIDMaxLen; overflow > 0 {
-		if overflow >= len(basename) {
+	// Two literal "-" separators plus the hash are the fixed overhead; the
+	// remaining budget is shared between mapping and basename.
+	budget := workflowIDMaxLen - (2 + workflowIDShortHashLen)
+	if len(mapping)+len(basename) > budget {
+		if len(mapping) >= budget {
+			mapping = mapping[:budget]
 			basename = ""
 		} else {
-			basename = basename[:len(basename)-overflow]
+			basename = basename[:budget-len(mapping)]
 		}
-
-		id = mapping + "-" + basename + suffix
 	}
 
-	return id
+	return mapping + "-" + basename + "-" + shortHash
 }
 
 // workflowIDSegmentInvalid matches any run of characters that should be
--- a/cmd/watcher/watcher.go
+++ b/cmd/watcher/watcher.go
@@ -10,6 +10,7 @@ import (
 	"log/slog"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


cmd/watcher/watcher.go
```diff
@@ -112,13 +112,88 @@ func newScanInstruments(reg prometheus.Registerer) (*scanInstruments, error) {
...
+	var prevUnderscore bool
+
+	for _, r := range s {
+		allowed := (r >= 'A' && r <= 'Z') ||
+			(r >= 'a' && r <= 'z') ||
+			(r >= '0' && r <= '9') ||
+			r == '.' || r == '_' || r == '-'
+		if !allowed {
+			r = '_'
+		}
```
solidDoWant: would a regex search and replace for non-compliant characters be simpler?

solidDoWant: Good call — done in 11a09e1. The whole helper collapses to a single regex `[^A-Za-z0-9.-]+` → `_`. I deliberately leave `_` out of the allowed class so runs of input underscores match the same regex and collapse for free, which removes the second pass / state machine entirely. All existing sanitization tests continue to pass.

cmd/watcher/watcher.go
```diff
@@ -112,13 +112,88 @@ func newScanInstruments(reg prometheus.Registerer) (*scanInstruments, error) {
...
+	id := mapping + "-" + basename + suffix
+	if overflow := len(id) - workflowIDMaxLen; overflow > 0 {
+		if overflow >= len(basename) {
+			basename = ""
+		} else {
+			basename = basename[:len(basename)-overflow]
+		}
+
+		id = mapping + "-" + basename + suffix
+	}
```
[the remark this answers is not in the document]
solidDoWant: Good catch — fixed in 22ca9e5. I reformulated the cap as a single shared `budget` between the mapping and basename segments rather than chaining two separate overflow checks: trim basename first so the operator-configured mapping name stays visible, then fall through to trimming mapping only when it alone exceeds the budget. Same end behaviour as your suggested patch, slightly less repetition. Added `TestWorkflowID_LengthCapTrimsMappingWhenItAloneOverflows` to lock in the mapping-overflow path.

solidDoWant:
Re the outside-diff suggestion to use `t.Context()` instead of `context.Background()` in the `TerminateWorkflow` cleanup at `cmd/watcher/integration_test.go:108-110` — keeping `context.Background()` here is intentional. Per `go doc testing.T.Context`: "Context returns a context that is canceled just before Cleanup-registered functions are called." So by the time the `t.Cleanup` callback fires, `t.Context()` is already canceled and `TerminateWorkflow` would return immediately with `context.Canceled` without actually terminating the test workflow. The CLAUDE.md rule "always use `t.Context()`" applies to the test body where the context is live; cleanup is the standard exception.

PR #235, opened 2026-05-26
fix(medialib): remove DownloadClientId from Sonarr scan; return ErrLibraryFileMismatch on size mismatch

solidDoWant:
## Summary

- **Fix 1 – Remove `DownloadClientId` from `DownloadedEpisodesScan`**: Including `DownloadClientId` caused Sonarr to scan the download client's content directory instead of the specified `Path`. For series packs this triggered a batch-import that produced a single `S01E01-E20` record overwriting all individual episode files. The fix drops `DownloadClientId` from the command entirely, then explicitly deletes the queue record via `DeleteQueueContext(RemoveFromClient: false)` after a successful scan — clearing `ImportPending` without disturbing the download client or affecting other in-progress episodes.
- **Fix 2 – Return `ErrLibraryFileMismatch` on post-check size mismatch (Sonarr + Radarr)**: When the race-recovery post-check finds that the arr service holds a file of a different size than expected, the previous code only logged a warning and fell through to retry. Since the arr service will never replace its existing file without operator intervention, retrying is futile. The fix returns `ErrLibraryFileMismatch` (wrapped), which the `Notify` activity surfaces as a `NonRetryableApplicationError` so Temporal stops burning retry budget.

## Test plan

- [x] `go test ./pkg/medialib/sonarr/...` — all `TestImportByFilePath*` cases pass, including new `wantQueueDeletedID` assertions and the renamed `ErrLibraryFileMismatch` case
- [x] `go test ./pkg/medialib/radarr/...` — `propagatesFileMismatch` check passes for size-mismatch case
- [x] `go test ./workflows/media/...` — new `TestNotify_LibraryFileMismatchIsNonRetryable` passes
- [x] `go test ./...` — full suite green
- [x] `make lint` — no issues

## Operator action required for existing failures

Workflows currently stuck on `ErrLibraryFileMismatch` (the `<Series Name>` `S01E01-E20` batch record):

1. In Sonarr, delete the file `<Series Name> - S01E01-E20- <First episode name>...<Second episode name> Bluray-1080p.mkv` from the library to unlink all 20 episodes.
2. Re-trigger the 26 failed workflow runs — they will now each import their correctly-named individual file.

### DIFF
--- a/pkg/medialib/medialib.go
+++ b/pkg/medialib/medialib.go
@@ -11,6 +11,14 @@ import (
 // ErrNotFound is returned when a media item is not found in the library.
 var ErrNotFound = errors.New("not found in library")
 
+// ErrLibraryFileMismatch is returned by ImportByFilePath when the arr service
+// already has a file for the requested media item but that file's size differs
+// from expectedSize. This indicates the arr service's existing file is from a
+// different import (not the file we just produced) and the service declined to
+// replace it. Retrying will not help; the operator must resolve the conflict in
+// the arr service before the workflow can succeed.
+var ErrLibraryFileMismatch = errors.New("library already has a different file for this media item")
+
 // MediaType identifies whether a media file is a movie or a TV show episode.
 type MediaType string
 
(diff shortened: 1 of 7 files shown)


solidDoWant:
closing this because it would break imports when the source file name has no clear mapping to a specific episode

